PR #19, opened 2023-01-20
Adding Community Detection App

jdr0887:
Initial stab at adding the CommunityDetection App.  Currently, only the form generation exists with a local store to capture the form values.


d2fong:
I think you mentioned this in the meeting already -- but I would agree the reasonable approach would be to create a form render component and handler specific to each algorithm from that algorithms response from the cdaps server: http://cdservice.cytoscape.org/cd/communitydetection/v1/algorithms

jdr0887:
I now have a first take on the onSubmit impl for the Run button.  There is no network sent in the post, but the response looks like the following.  What kind of file/data should I be sending to test that this works?  Here is what I am getting sans actual data in the request:

{"message": "Error requesting CommunityDetection",
"stackTrace": "com.fasterxml.jackson.databind.exc.MismatchedInputException.from(MismatchedInputException.java:59)com.fasterxml.jackson.databind.ObjectMapper._initForReading(ObjectMapper.java:4360)com.fasterxml.jackson.databind.ObjectMapper._readMapAndClose(ObjectMapper.java:4205)",
"threadId": "15412",
"description": "No content to map due to end-of-input  at [Source: (String)\"\"; line: 1, column: 0]",
"errorCode": null,
"timeStamp": "2023-01-24_14:22.55"}

jdr0887:
Closing to resync upstream changes, will create PR to capture downstream additions.

PR #21, opened 2023-02-02
Adding Community Detection App

jdr0887:


jdr0887 (commented):
@d2fong mind reviewing this PR?

src/components/CommunityDetection/RunCommunityDetectionFormDialog.tsx
```diff
@@ -0,0 +1,339 @@
...
+  InputLabel,
+  Select,
+  SelectChangeEvent,
+  TextField,
+  Tooltip
+} from '@mui/material'
+import * as React from 'react'
+import create from 'zustand'
+import { AlgorithmEntry } from './CommunityDetectionAlgorithmModel'
+
```
d2fong, line 23: import the network store and the workspace store:

```js
import { useNetworkStore } from '../../../store/NetworkStore'
import { useWorkspaceStore } from '../../../store/WorkspaceStore'
```

src/components/CommunityDetection/RunCommunityDetectionFormDialog.tsx
```diff
@@ -0,0 +1,339 @@
...
+  })
+}
+
+const formatKey = (algorithm: string, key: string): string => {
+  return `cd.${algorithm}.${key}`
+}
+
+export const RunCommunityDetectionFormDialog: React.FC = () => {
+  const { open } = useRunCommunityDetectionFormDialogStore()
+
```
d2fong, line 45: Get the current network id and the current network associated with that id:
```js
  const currentNetworkId: IdType = useWorkspaceStore(
    (state) => state.workspace.currentNetworkId,
  )
  const networks: Map<IdType, Network> = useNetworkStore(
    (state) => state.networks,
  )
```

d2fong: Refer to `Edge.ts`, `Node.ts` `GraphObject.ts` `Network.ts` for how to understand what `targetNetwork` is.  It is a simple model right now.
https://github.com/cytoscape/cytoscape-web/tree/development/src/models/NetworkModel

Here is a json example of it:

```js
{
id: 'networkId',
nodes: [{id: '0'}, {id: '1'}],
edges: [{id: 'e0', s: '0', t: '1'}]
}
```

d2fong:
Ok I left some comments in the 'files changed' tab.  Does this give you enough information to extract the data you need to send to the service?

d2fong:
Or do you need some node attributes/edge attributes data as well?

src/components/CommunityDetection/RunCommunityDetectionFormDialog.tsx
jdr0887: This is good.  I think the next step is to serialize a sif file into a Network instance.  Is there a util to do this?  Should I just read a sif file in & create a new Network?  Oh, wait, I see that there is a createFromSif function in the CyNetwork class.  Is that the right thing to use?

d2fong: It looks like the implementation of `createFromSif` is empty.  If you were planning to convert sif files to a new network I guess that would be the place to put your code.  How are you getting a sif file?  From a request?

jdr0887: I was planning on using one of the sif files from the Cytoscape desktop app.  There is a folder called sampleData with 3 sif files in it (BINDhuman.sif, BINDyeast.sif, and galFiltered.sif).  When I use currently hardcoded data, "1\t2\n1\t3\n2\t3\n", I get an error back from the CD service: 

`Received non zero exit code: 1 when running algorithm for task: 7d1c24d7-6933-4dea-8854-a3d1e37b3e24`

I suspect this is due to not using real data, but am unsure.

d2fong: I don't know the answers to those questions but I will bring it up Jing and Chris and they will reach out to you.

coleslaw481: Hidef does not appear to like that data. Made a ticket to see what the issue is.

PR #28, opened 2023-02-17
Renderer update

keiono:
View model and renderers are updated to use simple key-value pairs and direct data() mappers in Cytoscape.js.


keiono:
Update function still has issues (need to switch the selected network to reflect the edit in the vizmapper). I'll fix it.

src/models/VisualStyleModel/impl/MappingFunctionImpl.ts
```diff
@@ -239,7 +220,7 @@ export const typesCanBeMapped = (
 
   if (mappingType === 'continuous') {
     const vtIsNumber =
-      valueTypeName === 'integer' ||
+      valueTypeName === ValueTypeName.Integer ||
       valueTypeName === 'double' ||
       valueTypeName === 'long'
```
d2fong, line 225: Should all of these strings be replaced by their type enum?

d2fong (commented):
Had questions about two minor things in these changes.

PR #29, opened 2023-02-22
Update timing problems have been resolved

keiono:
There were unnecessary update calls in the renderer. They are removed and styles will be applied only once when the VS is modified.


keiono:
The rendering function will be blocked by a boolean value _isViewCreated_ and if that value is false, other style update function will not be called. That avoids duplicate "apply vs" functions during the view creation process. Also, useMemo hooks are added to avoid unnecessary updates after the initialization.

PR #30, opened 2023-02-22
use type enums instead of hard coded strings

d2fong:
Replace all instances(that I could find so far) of type specific strings with their equivalent enums constants.


keiono:
All of the changes are mechanical changes and looks good to me.

PR #31, opened 2023-02-23
When computing the a node/edge view, also check if the column has a default value when determining the row value to map

d2fong:
When getting the attribute value for a mapping in `computeView`, columns sometime contain a `defaultValue`.  If the row for a specific node/edge does not have a value, also check the column's defaultValue before deciding if the value is found or not.

### DIFF
--- a/src/models/VisualStyleModel/impl/visualStyle-utils.ts
+++ b/src/models/VisualStyleModel/impl/visualStyle-utils.ts
@@ -305,11 +305,13 @@ export const computeView = (
       pairs.set(name, bypass)
     } else if (mapping !== undefined) {
       // Mapping is available.
-      // TODO: compute mapping
       const attrName: string = mapping.attribute
-      const attributeValueAssigned: ValueType | undefined = row[attrName]
+      const column = columns.get(attrName)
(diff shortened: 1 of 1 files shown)


src/models/VisualStyleModel/impl/visualStyle-utils.ts
```diff
@@ -324,9 +326,7 @@ export const computeView = (
             `Mapping is defined, but Mapper for ${vp.name} is not found`,
           )
         }
-        const computedValue: VisualPropertyValueType = mapper(
-          attributeValueAssigned,
-        )
+        const computedValue: VisualPropertyValueType = mapper(attributeValue)
```
d2fong, line 329: Also the attributeValue could be an array of values.  Has validation already been performed or should we do a check here to make sure it is not an array?

d2fong:
defaultValue should not exist as a concept in cytoscape web.

src/models/VisualStyleModel/impl/visualStyle-utils.ts
keiono: ValueType includes lists, but yes, we need to handle list types in the implementation in the future versions.

PR #35, opened 2023-03-02
Remove the concept of 'defaultValue' for columns

d2fong:
Instead of storing defaultValue in the column model, apply the default value to each row if necessary and then do not store the default value in the column model any more.


keiono (approved):
Looks good to me.

PR #36, opened 2023-03-04
Selection bug

keiono:
Selection is now a part of Visual Style and has a special handler in the Cyjs mapping generator. Mouseover event is also handled and add underlay to show which element is pointed by the mouse pointer.


keiono:
Fixed to use default selection color for now since it is extremely rare to use custom selection color mapping.

PR #40, opened 2023-03-16
Export cx

d2fong:
- add conversion function for exporting cytoscape web models to cx
- add two menu options to allow the user to overwrite the current network to it's counterpart in ndex or create a copy as a seperate network
- compare modification times between the network in ndex and the modification time in cytoscape web to check if there is conflicting modifications. 
- if there is a conflict, present the user with a choice to confirm overwrite or to create a copy
- anytime the network is modified or copied, update the modification time locally in cytoscape web
-

### DIFF
--- a/src/components/ToolBar/DataMenu/CopyNetworkToNDExMenuItem.tsx
+++ b/src/components/ToolBar/DataMenu/CopyNetworkToNDExMenuItem.tsx
@@ -1,15 +1,5 @@
-import {
-  MenuItem,
-  Box,
-  Tooltip,
-  Dialog,
-  Button,
-  DialogTitle,
-  DialogContent,
-  DialogContentText,
-  DialogActions,
-} from '@mui/material'
-import { ReactElement, useContext, useState } from 'react'
+import { MenuItem, Box, Tooltip } from '@mui/material'
+import { ReactElement, useContext } from 'react'
 import { BaseMenuProps } from '../BaseMenuProps'
 
 // @ts-expect-error-next-line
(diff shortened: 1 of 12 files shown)


src/components/ToolBar/DataMenu/CopyNetworkToNDExMenuItem.tsx
```diff
@@ -0,0 +1,120 @@
...
+      const newNdexModificationTime = ndexSummary.modificationTime
+      updateSummary(currentNetworkId, {
+        modificationTime: newNdexModificationTime,
+      })
+
+      addNetworkToWorkspace(uuid as IdType)
+      // in the other places that this function is used, it seems that a setTimeout is required
+      // for it to work properly
+      // todo this should not be necessary
+      setTimeout(() => {
```
d2fong, line 82: anyway to fix this?

src/models/NetworkSummaryModel/NdexNetworkProperty.ts
```diff
@@ -1,9 +1,9 @@
 import { IdType } from '../IdType'
-import { ValueTypeName } from '../TableModel'
+import { ValueTypeName, ValueType } from '../TableModel'
 
 export interface NdexNetworkProperty {
   subNetworkId: IdType | null
-  value: string
+  value: ValueType
```
d2fong, line 6: I could be mistaken but it seemed that ndex network properties could be of any value type.  Can you confirm if these are really just strings?

d2fong:
- dont update old network sumary after creating copy
- making sure the newest loaded network is the current network
- optionally scroll or let the user see the latest current network in the list

d2fong:
- saving the network to the server/copy, write all the changes to the db

keiono:
This looks OK to me except one issue.
If user try to save someone else's network, the save menu should be disabled. Currently it fails silently with this error output in the console: 

image

We can fix this as a separate issue after merge.

PR #42, opened 2023-03-23
Fix bypass UI hover

d2fong:
- fix issues where hoveredElement and selectedNodes and selectedEdges were not being persisted across networkView updates
- implement applyHoverUpdate() in the cy.js renderer which applies hover styles when hoveredElement changes

### DIFF
--- a/src/components/NetworkPanel/CyjsRenderer/CyjsRenderer.tsx
+++ b/src/components/NetworkPanel/CyjsRenderer/CyjsRenderer.tsx
@@ -204,6 +204,19 @@ const CyjsRenderer = ({ network }: NetworkRendererProps): ReactElement => {
     console.log('#Time to  apply style: ', performance.now() - t1)
   }
 
+  const applyHoverUpdate = (): void => {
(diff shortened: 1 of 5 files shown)


src/components/NetworkPanel/CyjsRenderer/CyjsRenderer.tsx
```diff
@@ -204,17 +204,18 @@ const CyjsRenderer = ({ network }: NetworkRendererProps): ReactElement => {
...
-  //   if (cy != null) {
-  //     cy.nodes().removeClass('hovered')
-  //     cy.edges().removeClass('hovered')
-
-  //     if (hoveredElement != null) {
-  //       cy.getElementById(hoveredElement).addClass('hovered')
-  //     }
-  //   }
-  // }
+  const applyHoverUpdate = (): void => {
```
d2fong, line 207: This is currently implemented in the `CyjsRenderer` component.  Should this be moved to the `cyjs-utils` module?

Also do you have ideas on name should this function should have?

PR #45, opened 2023-03-31
Layout function with 3rd party layout support

keiono:
This branch is for the new layout architecture supporting both Cytoscape.js and 3rd party layout engines.

The Network object is passed to the layout engines, and actual View Model update will be done in the _afterLayout_ callback function.

This implementation is not optimal use for the memory, but OK for v1 use cases (networks with ~50k objects).


d2fong:
Reopening this PR to merge to development instead of master

PR #76, opened 2023-08-04
Table browser features

d2fong:
- Add table browser functions in the table store to create/edit/delete columns
- Show a button group of actions that can be performed by the user when they click a cell or a header
- Specific functions added: create/edit/delete column, apply value to column, apply value to selected nodes

In another PR I will try to make these menus show in a context menu via right click.  I was not able to get it working at the moment.


src/components/TableBrowser/TableBrowser.tsx
keiono, line 1: This is not urgent, but this file is getting bigger, and better to split smaller units later.

d2fong: Agreed.  I'll work on splitting it into more maintainable components tomorrow.

keiono (approved):
Some components are getting bigger and we may need to split smaller components later, but other than that it looks good to me.

PR #82, opened 2023-08-30
Network attribute fix

d2fong:


keiono (approved):
Checked the changes in the CX exporter part and it looks good to me.

PR #83, opened 2023-08-31
Network attribute fix

d2fong:
fix more issues related to network attributes


src/store/io/exportCX.ts
```diff
@@ -298,5 +291,7 @@ export const exportNetworkToCx2 = (
     ...aspects.map(({ key, aspect }) => ({ [key]: aspect })),
     { status },
   ]
+
+  console.log(cx)
```
keiono, line 295: Need to remove this (can be a long console output)

d2fong: right, thanks

PR #84, opened 2023-09-01
Fix issue related to network attributes not being exported as their r…

d2fong:
…espective data types


d2fong:
All properties are stored as  strings for easier editing and consistent data validation.  Once the properties need to be exported to cx, they will be converted to their proper types.

PR #85, opened 2023-09-05
additional fix for updating the table when the columns change

d2fong:
Accidentally left out a fix related to renaming tables.

Most of the changes here are auto formatting related, here is the link to the actual change:

https://github.com/cytoscape/cytoscape-web/compare/development...network-attribute-fix?expand=1#diff-48898f82a2d395175baf800e5f75a0d4d704f900ef4aef60ef8335d62dca8e76R149


keiono (approved):
This is not for this change, but eventually we need to move nodeTable and edgeTable to types (use the enum alternative pattern). Same for "node" and "edge."

PR #91, opened 2023-09-19
Table browser features

d2fong:
adds the following features to the table browser:
- resize columns (non-persisted)
- reorder columns (persisted)
- add menu option to select nodes/edges from the table browser 

Main change here is that the column model is a list of columns instead of a map lookup.  Would like feedback on this change specifically.


src/store/TableStore.ts
keiono, line 1: We can replace 'node' and 'edge' strings later by GraphObjectType (currently in my branch).

PR #98, opened 2023-10-27
Update network dialog

d2fong:
- when the user loads a network in the workspace in a url , let them know if there is a new version of it and prompt them to make a copy of their network and load the new version

### DIFF
--- a/src/components/AppShell.tsx
+++ b/src/components/AppShell.tsx
@@ -1,6 +1,6 @@
 import { Box } from '@mui/material'
-import { useState, ReactElement, useEffect, useRef, useContext } from 'react'
 import { Location, Outlet, useLocation, useNavigate } from 'react-router-dom'
+import { useState, ReactElement, useEffect, useRef, useContext } from 'react'
 import { useWorkspaceStore } from '../store/WorkspaceStore'
 import { getUiStateFromDb, getWorkspaceFromDb } from '../store/persist/db'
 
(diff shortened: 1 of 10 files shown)


src/components/AppShell.tsx
```diff
@@ -122,21 +142,63 @@ const AppShell = (): ReactElement => {
...
+          if (localNetworkOutdated) {
+            if (localNetworkModified) {
+              // local network and ndex network have been modified
+              // ask the user what they want to do
+              setShowDialog(true)
+            } else {
+              // the local network has not been modified but it has been modified on NDEx
+              // update the network silently
+              deleteNetwork(networkId)
+              await waitSeconds(2)
```
keiono, line 178: This (wait calls) should be resolved as a separate issue.

src/components/UpdateNetworkDialog.tsx
```diff
@@ -0,0 +1,120 @@
...
+          <>
+            <Button
+              variant="outlined"
+              disabled={!authenticated || loading}
+              onClick={async () => {
+                const parsed = parsePathName(location.pathname)
+                const { networkId } = parsed
+                setLoading(true)
+                deleteNetwork(networkId)
+                await waitSeconds(2)
```
keiono, line 90: This should be fixed as a separate problem, to.

keiono (approved):
There are two remaining issues (timing issue and custom "/network" endpoint problem, but OK to merge and fix those two later.

PR #105, opened 2023-11-17
fix bugs related to loading networks from  url

d2fong:
- routing logic used to assume that the network id exists in the workspace.  changed it to always add the network id to the workspace


keiono (approved):
Looks good to me. (Probably I forgot to add it when updated the routes)

PR #107, opened 2023-11-27
Dexie migrations

d2fong:
- adds `applyMigrations` function that applies migrations before the app is initialized
- each breaking model change should supply its own upgrade function, see commented example in migrations.ts


keiono (commented):
Need to test when the multiple view function is merged, but looks OK.

PR #108, opened 2023-11-27
Llm

d2fong:
Adds basic llm integration feature:
- add analysis menu item for running llm queries
- add menu item for viewing llm options with open ai api key input and a open ai model selector
- add template for gene set analysis


keiono:
I noticed Some usability issues:

1. Menu items are disabled by default, which is good, but will be enabled even if without the API key.
2. Some errors are invisible from users. For example, there is a 3RPM limit on my account but I cannot know it until I open the developer tool (see below)

image

keiono (approved):
One very minor point. When I select Run LLM Query, the menu does not close automatically. Once it is fixed, it is OK to merge.

PR #114, opened 2023-11-30
Clean cw 102

yihangx:
Move custom colors to models/ColorModel, so it can be reused. 

For the redundant code like:https://github.com/cytoscape/cytoscape-web/blob/2fa1a9034838f542fd2b54e4ee311c166af8d4fd/src/components/Vizmapper/Forms/VisualPropertyValueForm.tsx#L196 I do have different color pickers https://github.com/cytoscape/cytoscape-web/blob/2fa1a9034838f542fd2b54e4ee311c166af8d4fd/src/components/Vizmapper/VisualPropertyRender/Color.tsx#L67 for 'pickerRender' for 'color'. 

Any suggestions to simplify the code?


d2fong:
Ok this looks good.  We can chat more about the specifics of how to handle different color pickers in the meeting today.

PR #123, opened 2023-12-14
show error message when gene queries are empty

d2fong:
- disable regenerate response button when text field query is empty 
- show error messages when gene names not found from the hcx column values


d2fong:
I forgot a change to the llm pull request, I will close it and reopen it after I make the change and let you know

PR #130, opened 2024-01-29
Basic syncing tabs functionality

d2fong:
- use dexie observable to store timestamps everytime the cache changes
- when a user leaves and comes back to a tab, check if there have been changes via the timestamp and reload the page if there were changes.  
- reload the page to /networks/ if the current network was not found in workspace after switching tabs


keiono:
Tested basic sync functions including network deletion, and looks those work.

PR #131, opened 2024-02-03
Circle Packing view for browsing

keiono:
Basic functions for Circle Packing view have been implemented.

This branch is set to test with the "test" server and the following entry is used as the test case:

[45f5c116-8fda-11ee-96bc-005056aecf54](https://test.ndexbio.org/viewer/networks/45f5c116-8fda-11ee-96bc-005056aecf54)


keiono:
Please try with the test server, and then replace config.json and index.tsx (the line window.location.origin + '/cytoscape/silent-check-sso.html') to go back to the dev server settings.

PR #137, opened 2024-03-04
fix issue where certain keys would unfocus the search ndex bar

d2fong:
- stop event from propogating in the keydown press handler that propogated the events back up to the dropdown/other parent components


keiono (approved):
Looks OK to me and confirmed I can type all letters in search text box.

PR #138, opened 2024-03-05
fix bug where active network view would persist across current network id changes

d2fong:
Listen to current network id changes and update the active network view accordingly.


keiono (approved):
Looks good to me. (I think I forgot to implement this when add the activeNetwork prop.)

PR #139, opened 2024-03-06
Update cytoscape import

maxkfranz:
This updates the `cytoscape` import to use the default exports of the `cytoscape` package.

It was only luck that it worked before.

Before (bad):

```
import * as cytoscape from 'cytoscape'
```

After (good):

```
import cytoscape from 'cytoscape'
```

See https://github.com/cytoscape/cytoscape.js/issues/3217#issuecomment-1981627469

@GAOChengzhan @d2fong @keiono


d2fong (approved):
Thanks @maxkfranz.  Yes this change makes cytoscape-web work with the cy.js esm retooling.

PR #141, opened 2024-03-07
Fix CW-174

yihangx:
Fix CW-174


keiono (approved):
Confirmed the fix for the UI

PR #147, opened 2024-03-21
Cw 166

yihangx:


d2fong (changes requested):
Hi Yihang, 

Thanks for this work.  It looks good but I think there is a minor modification that can be made such that it does not break the rest of the code that deals with network summaries.  

Specifically, the exportCX function and all other places that deal with summaries would be broken.

I will describe the minor modification that can avoid breaking the other functions later today.

src/components/ToolBar/DataMenu/UploadNetworkMenuItem.tsx
```diff
@@ -82,20 +91,43 @@ export const UploadNetworkMenuItem = (props: BaseMenuProps): ReactElement => {
...
+            typeof item.networkAttributes[0].name === 'string'
+          ) {
+            localDescription = item.networkAttributes[0].description
+            break
           }
         }
+        const networkAttributeDeclarations =
+          getAttributeDeclarations(json).attributeDeclarations[0]
+            .networkAttributes
+        const networkAttributes = getNetworkAttributes(json)[0]
```
d2fong, line 117: `getAttributeDeclarations` and `getNetworkAttributes` are util functions from `models/CxModel/cx2-util.ts`.  These allow you to get various aspects of the cx and merge them to create proper `NdexNetworkProperty` objects for the summary.

d2fong (approved):

PR #149, opened 2024-04-17
Multi bypass value display

d2fong:
- when there is more than one unique bypass value selected, display a '?' in the box instead of the first bypass value.
- e.g. when there is a square and circle bypass value displayed, display '?' instead of square


keiono (approved):
Confirmed it displays "?" when multiple values are available in the bypass.

PR #150, opened 2024-04-19
users should not be able to create columns with an empty name

d2fong:


keiono (approved):
Confirmed the change to enable / disable the button.

PR #158, opened 2024-05-13
Table data loader

d2fong:
### Table data loader and table join
- add ui forms for creating a network from table data
- 

### Summary changes
- add a isNdex field to the summary model to determine if it is a ndex network or local network
- disable overwrite to ndex menu item if a network is local

### Default layout changes
- dont run a layout if summary.hasLayout is true
- if the network doesnt have a layout:
   - run grid layout if the network has more than the threshold of network elements
   - run hierarchical if the network is hcx and has less than the threshold of network elements
   - run force directed if the network is not hcx and has less than the threshold of network elements


keiono (approved):
The two remaining issues should be resolved in separate PR. Other than that, I confirmed basic functions work.

PR #159, opened 2024-05-14
Default Merge

GAOChengzhan:
Default Merge


src/features/MergeNetworks/components/MergeDialog.css
```diff
@@ -0,0 +1,73 @@
+.p-dropdown-panel {
```
d2fong, line 1: I think the current convention for this codebase is to not make separate css files for components, the css is just inline in the `sx` or `style` props.  

@keiono any thoughts on this?  

In my opinion, not a big deal for now though.

src/models/TableModel/impl/InMemoryTable.ts
```diff
@@ -143,6 +145,15 @@ export const insertRow = (
   table: Table,
   idRowPair: [IdType, Record<AttributeName, ValueType>],
 ): Table => {
+  // const tableCols = table.columns
```
d2fong, line 148: Is this code going to be used in the future?  If not I think this should be deleted.

src/models/TableModel/impl/InMemoryTable.ts
```diff
@@ -8,11 +8,13 @@ import { ValueTypeName } from '../ValueTypeName'
 import { CxValue } from '../../CxModel/Cx2/CxValue'
 import { AttributeDeclaration } from '../../CxModel/Cx2/CoreAspects/AttributeDeclarations'
 import { translateCXEdgeId } from '../../NetworkModel/impl/CyNetwork'
-export const createTable = (id: IdType): Table => ({
-  id,
-  columns: [],
-  rows: new Map<IdType, Record<AttributeName, ValueType>>(),
-})
+
+export const createTable = (id: IdType, colNames: Column[] = []): Table => (
```
d2fong, line 12: colNames should be renamed to `columns` if these are actually column objects.

src/features/MergeNetworks/utils/helper-functions.ts
```diff
@@ -0,0 +1,41 @@
+import { Pair } from "../models/DataInterfaceForMerge";
+
+//utility function to find index of a pair in a list
+export const findPairIndex = (pairs: Pair<string, string>[], uuid: string) => {
+    return pairs.findIndex(pair => pair[1] === uuid);
+};
+// Utility function for deep cloning objects that may contain arrays, maps, and functions
+export function deepClone<T>(obj: T): T {
```
d2fong, line 8: We use the package`lodash` which includes a deepClone function already.  Is there a specific case that this function you have here handles that lodash deepClone does not already handle?

src/features/MergeNetworks/models/Impl/CreateMergedNetworkWithView.ts
```diff
@@ -0,0 +1,102 @@
...
+    const newNetwork = mergedNetwork.network
+    const newNodeTable = mergedNetwork.nodeTable
+    const newEdgeTable = mergedNetwork.edgeTable
+
+    // Initialize new visual style and network view model
+    const newVisualStyle: VisualStyle = deepClone(visualStyle) || VisualStyleFn.createVisualStyle();
+    const newNetworkView: NetworkView = ViewModelFn.createViewModel(newNetwork)
+
+    await putNetworkSummaryToDb({
+        ownerUUID: toNetworkId,
```
d2fong, line 62: There are changes to the network summary in my table data loader branch.  

Specifically, a flag called `isNdex` needs to be set to false here.

I believe from past discussions it seemed reasonable that the result of merging networks produces a local network.

src/features/MergeNetworks/components/assets/cytoscape-3.ttf
d2fong, line 1: What is the purpose of this file?

d2fong (changes requested):
Mostly just need clarification on a few things.

src/features/MergeNetworks/components/assets/cytoscape-3.ttf
GAOChengzhan: This is a file for icons, originally I wanted to put the icons aside the button 'union', 'intersection' and 'difference' just like the cytoscape desktop, but I encountered some difficulty in showing them correctly, so for now I just keep it here.
image

src/features/MergeNetworks/models/Impl/CreateMergedNetworkWithView.ts
GAOChengzhan: Okay, I will add this property

src/models/TableModel/impl/InMemoryTable.ts
GAOChengzhan: okay

src/models/TableModel/impl/InMemoryTable.ts
GAOChengzhan: Yeah, I forget to delete these line.

src/features/MergeNetworks/utils/helper-functions.ts
GAOChengzhan: Okay, I will substitute the current deepClone with the one from `lodash`

src/features/MergeNetworks/components/MergeDialog.css
GAOChengzhan: Emmm, if I use cx or style, I think it will cause some code duplications since there are some components with the same class sharing the same style.

GAOChengzhan:
I commented out the 'Advanced options' in the dialog window just for the workshop since for now it is not so stable without the implementation of 'Type coercion'.

d2fong (approved):

PR #161, opened 2024-05-15
misc menu changes

d2fong:
- hide the help menu until proper content is available
- close the data menu after the user opens the table data loader forms


GAOChengzhan (approved):
Yeah, this makes sense. I will also add some code to close the menu item if a user starts a merge-network dialog.

PR #162, opened 2024-05-16
Solved a visual style and a toggle button bug

GAOChengzhan:
Solved two bugs:
1. A toggle button bug: [Jira ticket](https://cytoscape.atlassian.net/browse/CW-206)
2. A visual style bug: the visual style cannot be fetched from the store successfully if the network was not pre-loaded in the workspace, now I directly fetch the visual style information from the database


d2fong (approved):
Good work.

PR #164, opened 2024-05-17
Menu text changes

d2fong:
- moves the import network from file related menu items into a nested menu item
- handles cx2 and csv/tsv files
- uses the file extension to determine which import logic to run (cx2 or table data loader form)


keiono (approved):
Confirmed the new menus and worked as expected.

PR #165, opened 2024-05-17
ensure all nodes are initialized in the node table

d2fong:
- sometimes nodes may be in the nodes aspect but not in the node attr aspect
- initialze all nodes into the node table when converting cx


keiono (approved):

keiono:
Looks good to me. Confirmed the new initialization code.

PR #168, opened 2024-05-23
dont allow a user to open a collection from ndex

d2fong:
- grey out the network summary in the ndex modal if it is a collection
- a network summary is a collection if the subnetworkIds field is a non empty list


GAOChengzhan (approved):
Looks good to me!

keiono (approved):

PR #169, opened 2024-05-24
debug

GAOChengzhan:
It seems that the current dev branch has some bugs. There are mainly two kinds of errors:
1. Duplicate import -> I just delete one of them
2. 'NetworkView[]' is not assignable to parameter of type 'NetworkView' -> I replaced `viewModel`  with `viewModel[0]`

### DIFF
--- a/src/components/ToolBar/DataMenu/SaveWorkspaceToNDExOverwrite.tsx
+++ b/src/components/ToolBar/DataMenu/SaveWorkspaceToNDExOverwrite.tsx
@@ -71,8 +71,9 @@ export const SaveWorkspaceToNDExOverwriteMenuItem = (props: BaseMenuProps): Reac
     const summary = useNetworkSummaryStore.getState().summaries[networkId];
     const nodeTable = useTableStore.getState().tables[networkId].nodeTable;
     const edgeTable = useTableStore.getState().tables[networkId].edgeTable;
-    const viewModel = useViewModelStore.getState().viewModels[networkId];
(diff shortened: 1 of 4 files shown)


src/components/ToolBar/DataMenu/SaveWorkspaceToNDExOverwrite.tsx
```diff
@@ -80,7 +80,7 @@ export const SaveWorkspaceToNDExOverwriteMenuItem = (props: BaseMenuProps): Reac
       summary,
       nodeTable,
       edgeTable,
-      viewModel,
+      viewModel[0],
```
d2fong, line 83: Seems like @keiono said you should use `getViewModel` here instead.

GAOChengzhan:
Changed
```const viewModel = useViewModelStore.getState().viewModels[networkId];```
To 
```
const viewModel = useViewModelStore((state) => state.getViewModel(networkId))
if (viewModel === undefined) throw new Error("Cannot find view model for network!")
```

**Comment by GAOChengzhan** [comment]
Jing wants me let the **viewModel** could be `undefined` in the function `exportCX`, so I set the default nodes' **_x_** and **_y_** as 0 if **viewModel** is `undefined`
```
const nodes = network.nodes.map((node) => {
    const nodeRow = nodeTable.rows.get(node.id)

    return {
      id: parseInt(node.id),
      x: networkView?.nodeViews[node.id].x ?? 0,
      y: networkView?.nodeViews[node.id].y ?? 0,
      v: nodeRow,
    }
  })
```

keiono (approved):

PR #170, opened 2024-05-24
add network name field to the data table importer

d2fong:
- default name is the file name
- if the name already exists, append the first available number to it, e.g. file.csv1, file.csv2
- add a text input in the form to allow the user to change the name to something else


GAOChengzhan (approved):
Looks good to me.

keiono (approved):

PR #182, opened 2024-06-12
potential fix for the passthrough mapping flashing crash bug

d2fong:
### Context
- There exists a set of visual property values that are considered enums (node shape, edge line, visibility etc)
- All the current enum types except visibility are strings (visibility is a boolean)
- There is variation in how a user wants to passthrough map visibility:
   - 'TRUE', 'True', 'false', false 
 
This fixes a bug in the code that assumed that visibility would always be passed as a string (sometimes people pass a boolean column)

Just looking for comments to see if this is ok.  

There is another related bug to fix that stops the flickering from happening when each passthrough value throws an error.  Or whenever multiple values throw the same error again and again.

It may be more proper to not consider visibility as an enum type and then just have specific logic for visibility to be accepting of strings and booleans.

### Test Case:
1. Go to cytoscape web configured for dev
2. Load the network G2PT_Reactome
3. The network should no longer throw the error flashing screen

### DIFF
--- a/src/models/VisualStyleModel/impl/MapperFactory.ts
+++ b/src/models/VisualStyleModel/impl/MapperFactory.ts
@@ -12,7 +12,6 @@ import * as d3Scale from 'd3-scale'
 // import * as d3Color from 'd3-color'
 import { VisualPropertyValueTypeName } from '../VisualPropertyValueTypeName'
 
-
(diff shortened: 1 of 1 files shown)


keiono (approved):
Confirmed fix with the test data (G2PT_Reactome).

PR #184, opened 2024-06-13
fix bug where the search ndex input could not be focused

d2fong:
Add event handler in the parent modal element to prevent event propagation and prevent default.  When the user would click on the input it would trigger the menu items event handlers, overriding the input field event handler, preventing it from being focused.


keiono (approved):
Confirmed the fix by typing new search keyword in the text box.

PR #196, opened 2024-06-24
Solved A few issues in Network Merger

GAOChengzhan:
1. In the finding unique name function, when a network name already exists in the workspace, append “_n” to the name instead of “n”.  When a user does change the name to a duplicate name, change the text color to orange and add a message explaining the reason.  
image

2. In the default setting, name the new attribute “Matching.Attribute”  and rename the existing “Matching.Attribute” to “Matching.Attribute_n” if there is already one exists. Add checkers for duplicate attribute names. If the duplicates are detected, change the text to red and having a red error message somewhere in that dialog box.
image
image

3. When an error is detected, grey out the merge button and display the reason in the tooltip instead of popup an error message after the user clicks the merge button. 
image
image
 

4. Make the header of the 2 tables in the network selector section more obvious. 
image
 

5. Sort the list in the left window in alphabetical order (case insensitive).

6. Add tooltip to  “Matching.Attribute”  and also change the text color to make it more clear. 
image

### DIFF
--- a/src/components/ToolBar/ToolsMenu/MergeNetwork.tsx
+++ b/src/components/ToolBar/ToolsMenu/MergeNetwork.tsx
@@ -29,7 +29,7 @@ export const MergeNetwork = ({ handleClose }: BaseMenuProps): ReactElement => {
     const workSpaceNetworks: Pair<string, string>[] = networkIds.map((networkId) => {
         const networkName = networkSummaries[networkId]?.name
         return [networkName, networkId]
-    })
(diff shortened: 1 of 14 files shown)


GAOChengzhan:
Updates:
1. List Subheader UI
image


2. Rename the 'Matching.Attribute' to 'Matching.Attribute_n' if there is already such attribute name existing
image

GAOChengzhan:
Made the type conflict tooltip more detailed and updated other tooltips.
image

d2fong (approved):

PR #197, opened 2024-06-24
Original positions will be reset when external functions updates those

keiono:
Fixes inconsistent position problems by resetting those whenever external functions updates node positions.


GAOChengzhan:
I think this commit indeed solved the bug but a new problem also comes
When the user opens the layout tool and drags the slider, for example, they drag to 'x5', and then the corresponding node positions on this new scale are updated and recorded. If the user drags the slider again, the points on this new scale become the origin. So if the user drag back to 'x1', it would not come back to the initial state.

### Initial State:
image

### Drag to 'x5'
image

### Drag back to 'x1'
image

PR #202, opened 2024-06-26
CW-242

GAOChengzhan:
Ticket: [Jira CW-242](https://cytoscape.atlassian.net/browse/CW-242)

Since the ndexClient would not throw error by itself even when the returned network summaries' length is NOT equal to the requested ones. (see [src/store/hooks/useNdexNetworkSummary.ts](https://github.com/cytoscape/cytoscape-web/compare/CW-242?expand=1#diff-5394aab37718c35579ad094cc10635f869c5a997d783d5551380efd3ec91bbc5) **line 71**)

Now I added a if-statement to check this and if this happens(for example when the network is private and the user does not have permission), it will logs out an error. And [src/components/Workspace/WorkspaceEditor.tsx](https://github.com/cytoscape/cytoscape-web/compare/CW-242?expand=1#diff-dbc87ef8753984874b0e2ee607155749ba50b0b378ec8c83a804a8fa6fc85e27) **line 225** will delete all the networks that are requested to add but cannot be loaded.

### DIFF
--- a/src/components/TableBrowser/NetworkInfoPanel.tsx
+++ b/src/components/TableBrowser/NetworkInfoPanel.tsx
@@ -59,7 +59,6 @@ export default function NetworkInfoPanel(props: {
     (state) => state.summaries[currentNetworkId],
   )
   const properties = networkInfo?.properties ?? []
-  console.log(properties)
 
   const containsHtmlAnchor = (text: string) => {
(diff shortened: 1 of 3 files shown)


GAOChengzhan:
Added the user notification:
image

keiono (approved):

PR #206, opened 2024-07-02
add basic functionality to import networks from cx2 by url

d2fong:
- In the AppShell component, after the redirect logic runs:
   - Get the search parameters to see if there is a `import` key
   - Get the value of the `import` key and fetch the cx, convert the cx to cyweb models
   - Delete the `import` key so that the user doesnt run into the issue of importing the cx multiple times if they wanted to refresh the app


GAOChengzhan (approved):
Tested on my own machine, imported the network successfully.

PR #210, opened 2024-07-05
Solved CW-269

GAOChengzhan:
Jira: [CW-269](https://cytoscape.atlassian.net/browse/CW-269)

1. Corrected the condition of open dialog should depend on the variable **openDialog** instead of **open**. 
2. Only when the dialog is open then operate the function **fetchMyWorkspaces**. Otherwise, this function would be operated as long as the menu is open
3. Swapped the two templates, they seem to be reversed according to their naming.

(Please only focus on the updates under _src/components/Toolbar/DataMenu_, other updates are synced from other PR)

[CW-269]: https://cytoscape.atlassian.net/browse/CW-269?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/ToolBar/DataMenu/LoadWorkspaceDialog.tsx
+++ b/src/components/ToolBar/DataMenu/LoadWorkspaceDialog.tsx
@@ -50,10 +50,10 @@ export const LoadWorkspaceDialog: React.FC<{ open: boolean; handleClose: () => v
       return myWorkspaces;
     };
 
-    if (open) {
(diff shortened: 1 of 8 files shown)


yihangx (commented):
Reviewed.

yihangx (approved):

PR #216, opened 2024-07-13
Support for external server for root interaction network

keiono:
If the HCX metadata is correctly set, the root interaction network can be stored in different server. For example, even if the main hierarchy is stored in the test.ndexbio.org, the root interaction network can be uploaded to the public main server, ndexbio.org.

The attached is an example network. It has the root interaction network in the public server.

[sample_hcx.zip](https://github.com/user-attachments/files/16200631/sample_hcx.zip)


d2fong (approved):

keiono:
Test data is available here:

https://dev.ndexbio.org/index.html#/network/f3b67f31-43e5-11ef-a2cf-0242c246b7fb

PR #220, opened 2024-07-18
Image export

d2fong:
- add png, svg and pdf export
 - add module that registers cytoscape extensions (pdf, svg)
 - add menu item to register the export dialog


d2fong:
Yeah I will remove the unnecessary log statements.

d2fong:
Ok the log statements should be removed.

keiono (approved):

mikekucera:
Let me know if you run into any issues with the PDF export, including performance issues. I got the PDF extension to the point where it was basically working and needed more testing, and I haven't done much optimization yet.

PR #227, opened 2024-08-06
merge horizontal and vertical align types into node label position visual property

d2fong:


src/components/NetworkPanel/CyjsRenderer/cyjs-util.ts
```diff
@@ -47,6 +49,23 @@ export const createCyjsDataMapper = (vs: VisualStyle): CyjsDirectMapper[] => {
           },
         }
         cyStyle.push(selectedNodeMapping as CyjsDirectMapper)
+      } else if (vp.name === 'nodeLabelPosition') {
```
d2fong, line 52: add two different properties to the cyjs renderer when the vp is nodeLabelPosition

GAOChengzhan (approved):
It looks good to me. Yes, It may have conflicts with 'lock width and height' feature, but I think it's fine, I can change the code accordingly later.

PR #238, opened 2024-08-21
Vizmapper code cleanup

d2fong:
- Remove ~300 lines of unnecessary and complex code
- Move color picker code and logic from the VisualPropertyValueForm into the more appropriate ColorPicker component
- Improve style/layout of the color picker by centering the color swatches or picker and making the color picker content have a consistent height of 275 pixels


GAOChengzhan:
Hi @Dylan, I found a bug here

When you choose the network 'WP408 - Oxidative stress response - Homo sapiens' and then click `STYLE -> Nodes -> Shape (Map. column)`, it will trigger a bug
image

And another question is, we no longer use the scrollbar to let use choose the percentage value?
image

PR #240, opened 2024-08-26
Vizmapper code cleanup + cancel/confirm options for each visual property

d2fong:
- clean up vizmapper code (move color picker logic to color picker component instead of the parent visual property form)
- fix bug where an error would be thrown when invalid visual property values were given to the visual property value icon maps
- add cancel/confirm options to each visual property (preview when changing value, cancel or clickaway reverts to the old value, confirm keeps the previewed changes)
- change number input to behave the same way as all the other visual properties


d2fong (commented):
- fix the cancel confirm behaviour sync from parent value form to child value debounced form and vice versa
- fix stroke color and height lock box padding/position

keiono:
@d2fong If I select and cancel the default values, it works fine (go back to the original value), but if I use bypass, even if I cancel the value, still the new value will be applied. Is this a separate issue or a bug?

PR #242, opened 2024-08-26
Improve number validator

GAOChengzhan:
### Improved number validator

Jira: [CW-37](https://cytoscape.atlassian.net/browse/CW-37)

Added the check for the value range, now only the value in the range of **[0,1000)**  is considered as valid

Screenshot 2024-08-26 at 3 41 38 PM
image

[CW-37]: https://cytoscape.atlassian.net/browse/CW-37?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/assets/config.json
+++ b/src/assets/config.json
@@ -1,32 +1,22 @@
 {
-  "ndexBaseUrl": "dev.ndexbio.org",
+  "ndexBaseUrl": "dev1.ndexbio.org",
   "options": {
-    "debug": true
+   "debug": true
   },
   "keycloakConfig": {
-    "url": "https://dev.ndexbio.org/auth2",
-    "clientId": "localtestclient",
-    "realm": "ndex"
+   "url": "https://dev1.ndexbio.org/auth2",
+   "clientId": "cytoscapendex",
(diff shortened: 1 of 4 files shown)


d2fong (changes requested):
This conflicts with my change that makes the number input behave the same way as the other visual properties.  See the specific code here: https://github.com/cytoscape/cytoscape-web/pull/240/files#diff-78691b2e3aebd86847217bbf8f63c0e52f8aa4f9995902fc91ac2e0de6523575R37

After that change is merged, you can just use this component api to set the min/max values as well as whether the value is able to be negative or not.

https://mantine.dev/core/number-input/#min-and-max

GAOChengzhan:
Set max value for **opacity-related** and **rotation-related** visual properties while leaving other visual properties' max value empty for now(for example, size/font can be very large):
|Visual Property|Max Value|
|-------|-------|
|Opacity | 1|
|Rotation|360|

d2fong (approved):

PR #247, opened 2024-08-30
translate nodeshape

GAOChengzhan:
Jira: [CW-302](https://cytoscape.atlassian.net/browse/CW-302)
`Parallelogram` does not in exist _cytoscape.js_. Instead, it only exists `rhomboid` in _cytoscape.js_.  Therefore, I updated the convertor so that both 'parallelogram' and 'rhomboid' in cx2 are mapped to 'rhomboid' and it can be rendered correctly.

image

[CW-302]: https://cytoscape.atlassian.net/browse/CW-302?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/NetworkPanel/CyjsRenderer/cyjs-factory.ts
+++ b/src/components/NetworkPanel/CyjsRenderer/cyjs-factory.ts
@@ -4,10 +4,25 @@ import { Edge } from '../../../models/NetworkModel'
 import { ValueType } from '../../../models/TableModel'
 import { EdgeView, NodeView } from '../../../models/ViewModel'
 import {
-  VisualPropertyName, NodeVisualPropertyName, EdgeVisualPropertyName,
+  VisualPropertyName,
+  NodeVisualPropertyName,
+  EdgeVisualPropertyName,
   VisualPropertyValueType,
+  NodeShapeType,
 } from '../../../models/VisualStyleModel'
 import { VisualEditorProperties } from '../../../models/VisualStyleModel/VisualStyleOptions'
+
+export const NodeShapeMapping: Record<NodeShapeType, string> = {
+  [NodeShapeType.Parallelogram]: 'rhomboid',
+  [NodeShapeType.RoundRectangle]: 'roundrectangle',
+  [NodeShapeType.Triangle]: 'triangle',
+  [NodeShapeType.Diamond]: 'diamond',
+  [NodeShapeType.Octagon]: 'octagon',
+  [NodeShapeType.Hexagon]: 'hexagon',
+  [NodeShapeType.Ellipse]: 'ellipse',
+  [NodeShapeType.Rectangle]: 'rectangle',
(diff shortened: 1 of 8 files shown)


d2fong (changes requested):
This approach brings up some issues related to round tripping between cytoscape desktop. i.e. is it okay if the user imports cx2 with `parallelogram` values and then it exports back out to `rhomboid`?   

Another approach that keeps doesnt convert values is to just change the renderer to render `rhomboid` when value `parallelogram` is given.  We could talk about it in the meeting today.  

See here for an example renderring different shapes:
https://github.com/cytoscape/cytoscape-web/blob/development/src/components/NetworkPanel/CyjsRenderer/cyjs-util.ts#L235

GAOChengzhan:
> This approach brings up some issues related to round tripping between cytoscape desktop. i.e. is it okay if the user imports cx2 with `parallelogram` values and then it exports back out to `rhomboid`?
> 
> Another approach that doesnt convert values is to just change the renderer to render `rhomboid` when value `parallelogram` is given. We could talk about it in the meeting today.
> 
> See here for an example renderring different shapes: https://github.com/cytoscape/cytoscape-web/blob/development/src/components/NetworkPanel/CyjsRenderer/cyjs-util.ts#L235

Okay, got it. Let me update the code

src/components/NetworkPanel/CyjsRenderer/cyjs-factory.ts
```diff
@@ -25,15 +40,28 @@ export interface CyEdge {
...
       id: nv.id,
-      ...Object.fromEntries(nv.values.entries()),
+      ...Object.fromEntries(
+        Array.from(nv.values.entries()).map(([k, v]) => {
+          if (k === NodeVisualPropertyName.NodeShape) {
+            return [k, NodeShapeMapping[v as NodeShapeType]]
+          }
+          return [k, v]
+        }),
+      ),
```
GAOChengzhan, line 57: @d2fong Hi Dylan, I move the translate function here before giving the data to cytoscape.js.

src/components/NetworkPanel/CyjsRenderer/cyjs-util.ts
```diff
@@ -235,6 +237,8 @@ const updateCyObjects = <T extends View>(
             if (obj.data(key) === EdgeArrowShapeType.Arrow) {
               obj.data(key, EdgeArrowShapeType.Triangle)
             }
+          } else if (key === VisualPropertyName.NodeShape) {
+            obj.data(key, NodeShapeMapping[value as NodeShapeType])
```
GAOChengzhan, line 241: And also here

d2fong:
Ok looks good.

d2fong (approved):

PR #248, opened 2024-08-31
Fix for the performance issue when selecting nodes in large networks

keiono:
This should fix the performance issue when selecting nodes in large networks by adding new state to avoid unnecessary class removal process.

Cytoscape.js still display some warning messages in some situations, but it is from other type of problems (unnecessary fit?). But for now, this should be enough to fix the original performance issue.


d2fong (approved):
Great work.  That approach looks much more efficient.  Thanks.

PR #249, opened 2024-09-03
cytoscape desktop-like confirm/cancel

d2fong:
- clean up color vizmapper code
- implement basic confirm cancel button group
- user can only close the dialog via confirm/cancel
- remove 'preview' function that allowed user to preview the vizmapper changes before clicking confirm/cancel
- this approach may be converted to the preview/revert/confirm/cancel behaviour in the future if necessary


GAOChengzhan (approved):
Tested on my local machine, it runs smoothly. I find that the 'QuotaExceededError' also happens on dev branch(using localhost), but never seen this error on 'idekerlab.ndexbio.org/cytoscape' or 'test.ndexbio.org/cytoscape'.

d2fong:
Thanks.  That is interesting that you could reproduce the quota bug in local dev env.  I will try to see if I can do the same.

PR #251, opened 2024-09-06
fix bug where all table columns after the column to delete would be deleted

d2fong:
- splice only 1 element instead of all elements after start index


GAOChengzhan (approved):
Tested successfully with several networks.

PR #253, opened 2024-09-07
Sovled the bug: mismatch of degree and radian measure in Cy-web and cytoscape.js

GAOChengzhan:
1. Solved the bug: [Cytoscape uses Degree Measure while Cytoscape-web uses Radian Measure](https://cytoscape.atlassian.net/browse/CW-304)
2. Improve the code readability and efficiency by replacing **if - else if - else** structure with a map of handler functions.


d2fong (approved):
Great work!

PR #258, opened 2024-09-12
Help menu item

yihangx:


yihangx:
Adding "hidden cytoscape web tag" text in bug report form seems not possible because the Jira issue collector script may overwrite it. We should create a new ticket about this issue and contact Jira customer support.

PR #259, opened 2024-09-12
update node label position form

d2fong:
- users can choose 9 preset orientations, choose a text justification from a dropdown and set the x and y margin offsets


GAOChengzhan (approved):
It works perfectly now!

PR #261, opened 2024-09-12
derive node and edge counts from the model when loading cx networks from file

d2fong:
Derive node and edge counts from the generated network model instead of extracting them from the cx metadata when a network is loaded from a cx file


GAOChengzhan (approved):
Good job. This resolves the node/edge count bug.

PR #263, opened 2024-09-24
Fix selection bugs

d2fong:
- add additive/subtractive selection when the user holds down a modified key (cmd/ctrl/shift)
- fix various bugs related to tap events in cyJsRenderer.  
- The `renderNetwork` function was accessing state like `networkView` from outside the hook.  Those values would not be up to date and the render function would be rendering the wrong values.  Updated the `renderNetwork` hooks dependencies and changed the hook to call `renderNetwork` with them.

This fixes bugs such as:
- clicking the network after loading would render the wrong visibility values for nodes/edges 
- clicking the subnetwork edge causes all the edges to disappear

### DIFF
--- a/src/components/NetworkPanel/CyjsRenderer/CyjsRenderer.tsx
+++ b/src/components/NetworkPanel/CyjsRenderer/CyjsRenderer.tsx
@@ -161,149 +161,6 @@ const CyjsRenderer = ({
     exclusiveSelect(id, selectedNodes, selectedEdges)
   }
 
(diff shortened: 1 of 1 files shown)


d2fong:
I am missing knowledge/context about the selection feature in general, and in relation to the subnetwork viewer.  Should we also allow multiple selection for the subnetwork viewer?  As of right now the user can only click a single node in the subnetwork view.

keiono (approved):

PR #270, opened 2024-10-01
Set max and min for continuous mapping

GAOChengzhan:
### Refined the continuous mapping

- In the previous version, the mapping function would blindly do the mapping whether the attribute value is out of the domain or not.
- Now, the refined version would detect the attribute value and if it is out the defined domain, then apply the max/min vp value.


d2fong (commented):
@GAOChengzhan could a more simpler solution be found through the use of `clamp` from d3? https://d3js.org/d3-scale/linear#linear_clamp

GAOChengzhan:
> @GAOChengzhan could a more simpler solution be found through the use of `clamp` from d3? https://d3js.org/d3-scale/linear#linear_clamp

Okay, I used `Clamp` instead.

d2fong (approved):

PR #275, opened 2024-10-03
Fixes for leah demo

d2fong:
- fix issue where llm query panel could not scroll 
- fix issue where networks would not be fit to the screen after initial render

### DIFF
--- a/src/components/NetworkPanel/CyjsRenderer/CyjsRenderer.tsx
+++ b/src/components/NetworkPanel/CyjsRenderer/CyjsRenderer.tsx
@@ -694,9 +694,15 @@ const CyjsRenderer = ({
 
   useEffect(() => {
     if (cy !== null) {
+      renderNetwork(cy, networkView, displayMode, true)
+    }
+  }, [cy])
+
+  useEffect(() => {
+    if (cy !== null) {
       renderNetwork(cy, networkView, displayMode, false)
     }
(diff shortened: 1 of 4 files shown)


GAOChengzhan (approved):
Tested on my machine. It works fine.

src/components/NetworkPanel/CyjsRenderer/CyjsRenderer.tsx
```diff
@@ -692,11 +692,17 @@ const CyjsRenderer = ({
...
+      renderNetwork(cy, networkView, displayMode, true)
+    }
+  }, [cy])
+
   useEffect(() => {
     if (cy !== null) {
       renderNetwork(cy, networkView, displayMode, false)
     }
-  }, [cy, networkView, displayMode])
+  }, [networkView, displayMode])
```
GAOChengzhan, line 705: @keiono This may need your further review. Not sure whether it is proper to do so.

GAOChengzhan (approved):

src/components/NetworkPanel/CyjsRenderer/CyjsRenderer.tsx
keiono: Applying fit function only at the initialization should be OK.

keiono (approved):

PR #277, opened 2024-10-08
Fix hiview bugs

d2fong:
- disable fit and layout buttons for the circle packing view
- fix issue where the subsystem text would be overlayed on top of dialogs


GAOChengzhan (approved):
Tested on my computer. It's good.

PR #291, opened 2024-10-22
Add basic parameter UI

d2fong:
Renders a dialog box when the user clicks a service menu app item. 
Renders an node/edge column selector for each InputColumn model when the service input type is node or edge.
Render various parameter types for any service app.

Todo:
- Using validation regex
- Filtering columns by data type


GAOChengzhan:
Hi Dylan @d2fong. I tested `UpdateTablesExample`, it works.

Found two bugs:
1. The 'submit' button is enabled even when the workspace is empty.  If the submit is clicked, it will result in error.
image
image

2. The dropdown selection does not actually work in `iqueryenrichment`
image

GAOChengzhan:
Hi Dylan, I think this submit button is mistakenly disabled in the `updateTabelsExample`. It also checks whether the column exists, which is actually unnecessary
image

PR #299, opened 2024-10-24
CP renderer will be deleted in the manager instead of the panel

keiono:
This fixes the problem of showing unnecessary CP view panel tab. The renderer will be removed in the manager, so even if the panel is closed it hides the CP tab correctly.


GAOChengzhan (approved):
Tested on my machine, it works well.

PR #301, opened 2024-10-24
External Service 

GAOChengzhan:
|Update Tables Example| iQuery Enrichment|
| -------- | --------|
|image|image|

## Disabled case
image


GAOChengzhan:
I switch the `submit` button from left to right side so as to be aligned with other dialogs

keiono (approved):
Confirmed two example services works fine.

PR #302, opened 2024-10-26
Email Verification

GAOChengzhan:
#### Register
image

#### Dialog
image

### DIFF
--- a/src/bootstrap.tsx
+++ b/src/bootstrap.tsx
@@ -5,7 +5,7 @@ import './data-grid.css'
 import appConfig from './assets/config.json'
 import { AppConfigContext } from './AppConfigContext'
 import { App } from './App'
-
+import { EmailVerificationModal } from './components/EmailVerification'
 import ReactGA from 'react-ga4'
 
 // this allows immer to work with Map and Set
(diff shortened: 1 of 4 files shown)


GAOChengzhan:
The description in dialog and the button label may be discussed later to keep consistent across NDEx, Cy-web, Network Viewer, Iquery.

src/bootstrap.tsx
```diff
@@ -43,14 +50,62 @@ loadingMessage.textContent = 'Initializing Cytoscape. Please wait...'
...
 
 const keycloak = new Keycloak(keycloakConfig)
+
+const handleVerify = async () => {
+  await keycloak.loadUserProfile()
+  window.location.reload()
+}
+
+const handleCancel = () => {
+  keycloak.logout()
```
d2fong, line 60: I think there is a parameter here called redirectUri that is used in this logout function.  Depending on how cytoscape web is deployed e.g. test.ndexbio.org/cytoscape or web.cytsocape.org we might need to send a argument here.

src/bootstrap.tsx
```diff
@@ -43,14 +50,62 @@ loadingMessage.textContent = 'Initializing Cytoscape. Please wait...'
...
+}
+
+// Function to check if the user's email is verified
+const checkUserVerification = async () => {
+  try {
+    const ndexClient = new NDEx(appConfig.ndexBaseUrl)
+    ndexClient.setAuthToken(keycloak.token)
+    await ndexClient.getSignedInUser()
+    return {
+      isVerified: true,
```
d2fong, line 70: In my opinion, this function should always return a consistent object.  i.e. always return a object with `isVerified`, `userName` and `userEmail`, even if the the fields will not always be used by the calling function.

d2fong (commented):
Looks good, we just need to see if we need we need to use a redirectUri when keycloak logs out across the various cyweb deployments.

d2fong (approved):

PR #309, opened 2024-10-30
Merge service apps under parent menus if they share the same parent

d2fong:


GAOChengzhan (approved):
Tested and it works well.

PR #310, opened 2024-10-30
Replaced getSignedInUser() with signInFromIdToken()

GAOChengzhan:
#### Login 
1. Used signInFromIdToken() instead
2. Make suer the app would not hit the route and only show the notification dialog when the user's email is not verified
image

#### Save to NDEx Menu Item
Set the tooltip for case when a user is not loggedin

### DIFF
--- a/src/bootstrap.tsx
+++ b/src/bootstrap.tsx
@@ -52,7 +52,6 @@ document.body.appendChild(loadingMessage)
 const keycloak = new Keycloak(keycloakConfig)
 
 const handleVerify = async () => {
-  await keycloak.loadUserProfile()
   window.location.reload()
 }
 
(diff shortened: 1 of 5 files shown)


src/bootstrap.tsx
```diff
@@ -75,11 +88,11 @@ const checkUserVerification = async () => {
       e.status === 401 &&
       e.response?.data?.errorCode === 'NDEx_User_Account_Not_Verified'
     ) {
-      const userInfo: UserInfo = (await keycloak.loadUserInfo()) as UserInfo
+      const userInfo = parseMessage(e.response?.data?.message)
```
d2fong, line 91: @GAOChengzhan How come we no longer call loadUserInfo here?

GAOChengzhan: Oh, Jing wants me to get the info from the error message directly instead of calling this function which is not up-to-date

d2fong (approved):

PR #314, opened 2024-11-06
Fix table browser selection

d2fong:
- Fix issue where multiple rows couldnt be selected with the modifier key pressed.
- Fixed issue where selected row states would persist after performing the select nodes from selected rows operation.
- Fixed issue where selected row states would persist after changing between the node/edge table or loading a different network.
- Fixed issue where selecting edges from selected rows would continue to select nodes instead.
- Removed checkboxes as they didn't function properly and conflicted with the cell click event.

### DIFF
--- a/src/components/TableBrowser/TableBrowser.tsx
+++ b/src/components/TableBrowser/TableBrowser.tsx
@@ -171,6 +171,14 @@ export default function TableBrowser(props: {
     columns: CompactSelection.empty(),
     rows: CompactSelection.empty(),
   })
+
+  // Reset selection after switching table tabs
+  useEffect(() => {
+    setSelection({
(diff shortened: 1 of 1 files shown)


GAOChengzhan:
Hi Dylan @d2fong , Nice job on the fix! 
And I'm still encountering a couple of issues with this branch:

- It seems I can no longer edit the table cells. Was this functionality intentionally disabled?
- When I press and hold the 'Shift' key to select multiple lines, it selects lines [M, N-1] instead of [M, N]. For example, if I click on line M first and then line N, it only selects up to N-1.
- Another minor thing is that I cannot deselect the lines/columns anymore if I click blank areas in the table

GAOChengzhan (approved):
It works smoothly.

PR #329, opened 2024-11-14
Continuous mapping values fix

d2fong:
- Add feature to let users change the color/number mapping of values that fall outside the continuous mapping domain


d2fong:
Noticed something else that needed to be fixed.  Closing for now.

PR #342, opened 2024-11-22
Fix the position of the min and max handles in the continuous mapping…

d2fong:
… forms

- Users can't freely drag the min and max handles in either of the continuous mapping forms.


GAOChengzhan:
Hi Dylan, everything is good except the following bug:

In this case the `min` is `1` while the `max` is `4`

image


When adding a new handle and opening the input form, it prevents the user from entering out of the range value, but if I quit the input form and I can still click the 'Add Handle' button anyway, it can still add the handle successfully. I guess the  'Add Handle' button also needs to validate the value 

image
image

GAOChengzhan:
From today's meeting, there are a couple of advice:
1. These items are out of the gray background, maybe it's better to add more paddings to the gray background to make it wrap the inner components
Screenshot 2024-11-26 at 12 10 34 PM

2. The 'out-of-bound' input value for number mapping is not so intuitive, maybe we should apply the same style in color mapping, namely lift it up to just below the two end handles
Screenshot 2024-11-26 at 12 25 36 PM

PR #343, opened 2024-11-23
Simple Fix for CW-376 (missing edges)

keiono:
I've tried several ideas for show/hide edge features for filtering, but it requires major change in the renderer, so for this round, I'll keep the same mechanism, but use the normal selection mode whenever filters are not available for the interaction network.


GAOChengzhan (approved):
@keiono So this is resolved by setting every subnetwork using a different config (previously they shared the default config)?

PR #362, opened 2024-12-05
disable open in cytoscape button in safari

yihangx:


d2fong (approved):

GAOChengzhan:
Hi Yihang, I think you missed another 'Open in Cytoscape' button in 'Data-> Open Copy of Current Network in Cytoscape' in the navigation bar (the above changes only fix the one in the floating toolbar). And since two components try to detect whether it is on safari, an improvement for the code could be to detect this in the root app and pass the parameter to the components.

PR #369, opened 2024-12-09
Fix various issues in the continuous mapping form

d2fong:
- Rework gt/lt controls in the color and number form
- Fix issue where the color schemes were not applied properly (cold/hot values swapped)
- Fix issue where middle handle was not positioned in the mid point when creating a continuous mapping


GAOChengzhan (approved):
Tested and it works well

PR #372, opened 2024-12-17
fix open in cytoscape menuitem in data menu

yihangx:
disable this menuitem when browser is safari

### DIFF
--- a/src/components/FloatingToolBar/OpenInCytoscapeButton.tsx
+++ b/src/components/FloatingToolBar/OpenInCytoscapeButton.tsx
@@ -126,16 +126,18 @@ export const OpenInCytoscapeButton = ({
 
   return (
     <>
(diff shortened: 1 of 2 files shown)


GAOChengzhan:
Hi Yihang, thanks for your work. I tested on Safari and it works well.
Here are some advice for improvements:
1. Consistency in tooltips
I think it's better to also add a tooltip on the floating toolbar saying "This feature is not available on Safari"

|Data Menu|Floating Toolbat|
|------|------|
|image|image|

2. Consistency in snack bar
I noticed the two ways of 'opening network in Cytoscape' trigger different snack bars. I think it is better to keep them in the same style.

|Data Menu|Floating Toolbat|
|------|------|
|image|Screenshot 2024-12-19 at 12 35 56 PM|

PR #376, opened 2024-12-18
Fix opaque store bug & Add ndex summary checker & Improve cx2 exporter

GAOChengzhan:
This PR 
- Added an NDEx summary checker after save/update the network to NDEx sothat we can ensure NDEx accepts the network (See `src/utils/ndex-utils.ts`).
- Changed the cx2 exporter:  added the **attribute type**  to the mapping's definition when the attribute does not exist in the node/edge table (See `src/store/io/exportCX.ts`).

    (The above two changes resolved the [CW-437](https://cytoscape.atlassian.net/browse/CW-437) since its caused by the exported cx2 file of U2OS network is not accepted by NDEx)

- Added `getOpaqueAspect` in `getCachedData` function to resolve [CW-452](https://cytoscape.atlassian.net/browse/CW-452) (See `src/utils/cx-utils.ts`). Otherwise getting all cached data will never be fulfilled since the OpaqueAspect data is always missing. Then loading a local cx2 file will always trigger a `get` call to NDEx and receive 404 since obviously there is no such network in NDEx. @yihangx 

- Resolved the bug in **OpaqueStore**. 
    The screenshot of the error:
image 
@d2fong `Zustand` store is using `immer` as middleware, which manipulates the state using a Proxy to enable immutable state updates. The issue arises when **asynchronous operations**, like `putOpaqueAspectsToDb`, `deleteOpaqueAspectsFromDb`, access state after the state proxy has been revoked and the state is passed directly to asynchronous functions or persisted while still being a proxy (See `src/store/OpaqueAspectStore.ts`). This bug further causes the mismatch of the URL and current network ID, now [CW-462](https://cytoscape.atlassian.net/browse/CW-462) and [CW-454](https://cytoscape.atlassian.net/browse/CW-454) @yihangx are also resolved. 

    An example:
```
addAll: (networkId: IdType, aspects: Record<string, any[]>[]) => {
      set((state) => {
        if (!state.opaqueAspects[networkId]) {
          state.opaqueAspects[networkId] = {}
        }
        aspects.forEach((aspect) => {
          const [aspectName, aspectData] = Object.entries(aspect)[0]
          state.opaqueAspects[networkId][aspectName] = aspectData
        })
        void putOpaqueAspectsToDb(
          networkId,
          state.opaqueAspects[networkId],  //the state is passed directly to asynchronous functions 
        ).then(() => {
          console.debug('DB Update: opaque aspects store')
        })
        return state
      })
    },
```

[CW-437]: https://cytoscape.atlassian.net/browse/CW-437?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CW-452]: https://cytoscape.atlassian.net/browse/CW-452?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CW-462]: https://cytoscape.atlassian.net/browse/CW-462?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/ToolBar/DataMenu/CopyNetworkToNDExMenuItem.tsx
+++ b/src/components/ToolBar/DataMenu/CopyNetworkToNDExMenuItem.tsx
@@ -23,6 +23,7 @@ import { HcxValidationSaveDialog } from '../../../features/HierarchyViewer/compo
 import { NetworkView } from '../../../models/ViewModel'
 import { useUiStateStore } from '../../../store/UiStateStore'
 import { useOpaqueAspectStore } from '../../../store/OpaqueAspectStore'
+import { saveCopyToNDEx as saveNetworkCopy } from '../../../utils/ndex-utils'
 
 export const CopyNetworkToNDExMenuItem = (
   props: BaseMenuProps,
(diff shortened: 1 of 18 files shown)


d2fong (approved):
Tested the branch and it looks like the bugs are fixed.

PR #377, opened 2024-12-19
fix cannot save local network to workspace

yihangx:
fix CW-449 https://cytoscape.atlassian.net/browse/CW-449


GAOChengzhan:
Hi Yihang, 

Great work on this fix! However, I’d hesitate to move forward just yet as I still have some concerns. I’ve reached out to Jing on Slack and mentioned you there as well.

In addition, I encountered a bug:

There are two "Save Workspace" buttons—one for creating a new entry and another for updating the current workspace. It seems you only updated the former. Could you please make a similar change in the latter component? You can find it in `src/components/ToolBar/DataMenu/SaveWorkspaceToNDExOverwrite.tsx`.

Thanks!

yihangx:
Thanks! Sure, we can discuss this with Jing.

GAO Chengzhan ***@***.***>於 2024年12月19日 週四，13:23寫道：

src/components/ToolBar/DataMenu/SaveWorkspaceToNDEx.tsx
```diff
@@ -128,17 +128,16 @@ export const SaveWorkspaceToNDExMenuItem = (
       )
 
       const workspace = await getWorkspaceFromDb(currentWorkspaceId)
+      const onlyNdexNetworkIds = workspace.networkIds.filter(id => summaries[id]?.isNdex !== false);
```
GAOChengzhan, line 131: I think it is better to change `id => summaries[id]?.isNdex !== false` to `id => summaries[id]?.isNdex ===true`.

The reason for that is: 

- `summaries[id]?.isNdex === true `explicitly checks that the value of isNdex is true (i.e., the boolean true).
- `summaries[id]?.isNdex !== false `is broader and includes any value that is not strictly `false`, such as `undefined`, `null`

GAOChengzhan:
@yihangx Hi Yihang, are you still working on this or it is finished?

yihangx:
> @yihangx Hi Yihang, are you still working on this or it is finished?

It is finished.

src/components/ToolBar/DataMenu/SaveWorkspaceToNDExOverwrite.tsx
```diff
@@ -95,36 +100,53 @@ export const SaveWorkspaceToNDExOverwriteMenuItem = (
...
+              const ndexClient = new NDEx(ndexBaseUrl)
+              const token = await getToken()
+              ndexClient.setAuthToken(token)
+              const myNetworks = await ndexClient.getAccountPageNetworks(0, 1000)
+              return myNetworks
+            }
+      const myNetworks = await fetchMyNetworks();
+      const onlyNdexNetworkIds = myNetworks.filter((network: { externalId: string }) =>
+              workspace.networkIds.includes(network.externalId)
+            );
```
GAOChengzhan, line 113: First of all, this function is repeatedly defined in both 'SaveWorkspaceToNDExOverwrite.tsx' and 'SaveWorkspaceToNDEx'. It is better to avoid such code duplication.

And I think it is unnecessary to fetch the networks from NDEx here since the info of the current workspace stored in the indexed DB is all what we need. You can refer to my update [here](https://github.com/cytoscape/cytoscape-web/pull/394/files#diff-07d4d50565e8a0ce944dbf8e40a2587b062b1e50b0d860e975db7a7ef0b49f15R147).

src/components/ToolBar/DataMenu/SaveWorkspaceToNDExOverwrite.tsx
```diff
@@ -95,36 +100,53 @@ export const SaveWorkspaceToNDExOverwriteMenuItem = (
...
         setId(uuid)
       }
-
+      
+      const onlyLocalNetworks = workspace.networkIds.filter((id) =>
+        !myNetworks.some((network: { externalId: string }) => network.externalId === id)
+      );
+      
+      deleteNetwork(onlyLocalNetworks)
+
```
GAOChengzhan, line 135: I think it's better to move this **delete** operation to the function `saveCopyToNDEx` in the file ['src/utils/ndex-utils.ts'](saveCopyToNDEx), because not only two 'save workspace' menu items need to call this function but also the two 'save network' menu items.

GAOChengzhan:
Hi Yihang @yihangx , I made some comments above. And as this PR is a little bit messy (I guess it's the GitHub problem, it can not clearly show the diff between two branches and I have to manually check the 'compare' (like [here](https://github.com/cytoscape/cytoscape-web/compare/development...fix-save-workspace-with-local-networks) to see what changes you've made.), so I started a new [PR](https://github.com/cytoscape/cytoscape-web/pull/394)  which cleans your code and made some adjustments.

GAOChengzhan:
I will close this PR as [#394](https://github.com/cytoscape/cytoscape-web/pull/394) would solve this bug.

GAOChengzhan:
I will close this PR as [#394](https://github.com/cytoscape/cytoscape-web/pull/394) would solve this bug.

PR #383, opened 2025-01-06
Fix bugs related to cx2 file upload

d2fong:
- Use safe checks for accessing network attributes.
- Remove unsafe logic for accessing name and description from cx.
- Add a check in the node position aspect to see if the network already has a layout and set the flag in the network summary accordingly.
- If the name doesn't exist in the file attributes, use the same logic for creating a name for the network when a user creates a network from a table.


GAOChengzhan (approved):
LGTM

PR #386, opened 2025-01-08
Fix issue where users couldn't move the selected cell with arrow key …

d2fong:
…presses

To test:
1. Open a network
2. Select a cell in the table browser
3. Press the arrow keys to move the selected cell


GAOChengzhan (approved):
Tested and it works perfectly

PR #388, opened 2025-01-08
Resolve 'cannot delete all service app' bug

GAOChengzhan:


GAOChengzhan:
@keiono Hi Kei, noticed you just created a PR that would also resolve the bug, so I just close this one.

PR #391, opened 2025-01-10
Fix issue where the ValueTypeName.Double enum would not have a corres…

d2fong:
…ponding valid column assignment type array returned in the table data loader

refs CW-444

To test:
- refer to the instructions in CW-444


GAOChengzhan (approved):
Tested the code and it resolves the bug.

PR #394, opened 2025-01-12
Delete the original network when saving copy of a network

GAOChengzhan:
|To test, follow the steps |
|-----|
|1. Import one or more cx2 file to the current workspace|
|image|
|2. Click "Save Workspace as" in the **Data** menu and enter a unique workspace name|
|image|
|3. Click 'Save'|
|image|
|4. Check the workspace status in "Data -> Open Workspace from NDEx..."|
|Screenshot 2025-01-12 at 3 35 48 PM|

* This PR resolves this [ticket](https://cytoscape.atlassian.net/browse/CW-449)

### DIFF
--- a/src/store/io/exportCX.ts
+++ b/src/store/io/exportCX.ts
@@ -185,11 +185,6 @@ export const exportNetworkToCx2 = (
 
   const networkAttributes: any = [{}]
 
-  if (networkName != null || summary.name != null) {
-    networkAttributeDeclarations.name = { d: 'string' }
-    networkAttributes[0].name = networkName ?? summary.name
-  }
-
   summary.properties.forEach((property) => {
     networkAttributes[0][property.predicateString] =
       isListType(property.dataType) && !Array.isArray(property.value)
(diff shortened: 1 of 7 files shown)


src/store/io/exportCX.ts
```diff
@@ -200,6 +195,11 @@ export const exportNetworkToCx2 = (
         : property.value
   })
 
+  if (networkName != null || summary.name != null) {
+    networkAttributeDeclarations.name = { d: 'string' }
+    networkAttributes[0].name = networkName ?? summary.name
+  }
+
```
GAOChengzhan, line 202: This block of code needs to be moved below the `summary.properties.forEach((property) => {...` because the network `name` property in `networkAttributes` would be overwritten otherwise.

src/utils/ndex-utils.ts
```diff
@@ -117,6 +118,7 @@ export const saveCopyToNDEx = async (
   viewModel?: NetworkView,
   visualStyleOptions?: VisualStyleOptions,
   opaqueAspect?: OpaqueAspects,
+  deleteOriginal?: boolean,
```
GAOChengzhan, line 121: There are two use cases for creating a copy of the current network: deleting the original network or keeping it intact. The latter use case will be applied in this [ticket](https://cytoscape.atlassian.net/browse/CW-478)

d2fong (approved):

PR #396, opened 2025-01-16
Improve User Experience on Data Menu

GAOChengzhan:
#### 1. Menu Item Rename
- If a menu item name ends with `...`, it means it will open a dialog.
- Get rid of some words like `current` or `network` to be more concise
image

#### 2.Disable certain menu items when the workspace is empty
- 'Open workspace' is also disabled when the user is anonymous
image

#### 3. Change the behavior of `Save Network to NDEx` (previously called `Save Current Network to NDEx (overwrite)`)
- Enable it to save the `local` networks (previously it could not do this)
- It will create a copy for the local network, save it to NDEx and then delete the origin one
    - The behavior is similar to `Save Copy to NDEx` (previously called `Save Current Network to NDEx as...`). The only difference is that the latter would not delete the original one
    - When there are local networks in the workspace, `Saving workspace` operation  (no matter creating a new workspace entry or updating an existing one) will adopt the `save copy and delete original one` strategy

#### 4. Make tooltips on the right so that it would not block the button below it 
|After change| Before change|
|-----|-----|
|image|image|

### DIFF
--- a/src/components/ToolBar/DataMenu/CopyNetworkToNDExMenuItem.tsx
+++ b/src/components/ToolBar/DataMenu/CopyNetworkToNDExMenuItem.tsx
@@ -104,7 +104,7 @@ export const CopyNetworkToNDExMenuItem = (
         viewModel,
         visualStyleOptions,
         opaqueAspects,
-        true,
+        false, // keep the original network
       )
       setCurrentNetworkId(uuid as IdType)
       addMessage({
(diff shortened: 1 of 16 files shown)


GAOChengzhan:
Ticket: CW-478](https://cytoscape.atlassian.net/browse/CW-478)

d2fong (approved):

PR #398, opened 2025-01-18
Save App to Workspace

GAOChengzhan:
Ticket: [CW-480](https://cytoscape.atlassian.net/browse/CW-480)

#### 1. Pick a workspace, then add a few certain service apps and activate certain app(s). (Remember what you added/activated)
image

#### 2. Save the current workspace
image

#### 5. Open the `App -> Manage Apps` to see if the app setting are exactly the same as what you set

[CW-480]: https://cytoscape.atlassian.net/browse/CW-480?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/ToolBar/DataMenu/LoadWorkspaceDialog.tsx
+++ b/src/components/ToolBar/DataMenu/LoadWorkspaceDialog.tsx
@@ -22,7 +22,10 @@ import { AppConfigContext } from '../../../AppConfigContext'
 import { useCredentialStore } from '../../../store/CredentialStore'
 import { useWorkspaceStore } from '../../../store/WorkspaceStore'
 import { fetchMyWorkspaces } from '../../../utils/ndex-utils'
-import { Workspace } from '../../../models'
+import { MessageSeverity, Workspace } from '../../../models'
+import { useAppStore } from '../../../store/AppStore'
+import { useMessageStore } from '../../../store/MessageStore'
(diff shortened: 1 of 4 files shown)


src/components/ToolBar/DataMenu/LoadWorkspaceDialog.tsx
```diff
@@ -85,14 +91,56 @@ export const LoadWorkspaceDialog: React.FC<{
...
+          currentServiceApps.forEach((serviceAppKey) => {
+            if (!activeServiceApps.has(serviceAppKey)) {
+              removeServiceApp(serviceAppKey)
+            }
+          })
+          activeServiceApps.forEach((serviceAppKey) => {
+            if (!currentServiceApps.has(serviceAppKey as string)) {
+              addServiceApp(serviceAppKey as string)
+            }
+          })
```
GAOChengzhan, line 127: When loading a remote workspace, what I do here for either app or service app is:
- Check the current app list
- Deactivate/Remove the apps that are in the current local workspace but not stored in the remote workspace being loaded
- Activate/Add the apps that are stored in the remote workspace being loaded but not in the current local workspace

d2fong (approved):

PR #400, opened 2025-01-21
Render annotations for cy.js networks

d2fong:
- Read only annotation display for cytoscape.js networks
- Use the same package version as the ndex angular app for displaying annotations
- Remove previous layers on each render

### DIFF
--- a/src/components/NetworkPanel/CyjsRenderer/CyjsRenderer.tsx
+++ b/src/components/NetworkPanel/CyjsRenderer/CyjsRenderer.tsx
@@ -7,6 +7,10 @@ import Cytoscape, {
   NodeSingular,
   SingularElementArgument,
 } from 'cytoscape'
+// @ts-expect-error-next-line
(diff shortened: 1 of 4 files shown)


src/components/NetworkPanel/CyjsRenderer/CyjsRenderer.tsx
```diff
@@ -30,9 +34,10 @@ import {
   Orientation,
   PaperSize,
 } from '../../ToolBar/DataMenu/ExportNetworkToImage/PdfExportForm'
+import { useNetworkSummaryStore } from '../../../store/NetworkSummaryStore'
+import { CX_ANNOTATIONS_KEY } from './annotations/CyAnnotation'
```
GAOChengzhan, line 38: @d2fong Hi Dylan,

I got an error for this line when building the app and the netify also fails to build:
Screenshot 2025-01-21 at 5 31 18 PM

GAOChengzhan (approved):
Tested on localhost and it works well. And the code looks good to me.

PR #403, opened 2025-01-22
create a new workspace id in the local if the deleted NDEx workspace UUID is same as the local workspace UUID

yihangx:
https://cytoscape.atlassian.net/jira/software/projects/CW/issues/CW-482

### DIFF
--- a/src/components/ToolBar/DataMenu/LoadWorkspaceDialog.tsx
+++ b/src/components/ToolBar/DataMenu/LoadWorkspaceDialog.tsx
@@ -1,5 +1,6 @@
 /* eslint-disable @typescript-eslint/strict-boolean-expressions */
 import React, { ReactElement, useState, useEffect, useContext } from 'react'
+import { v4 as uuidv4 } from 'uuid'
 import {
   Dialog,
   DialogTitle,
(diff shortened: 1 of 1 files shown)


GAOChengzhan (changes requested):
@yihangx Please read the ticket carefully, this is unrelated to the bug.

The root of this is because in "Save Workspace Overwrite" menu item, it checks the workspace exist or not based on **the indexed DB** instead of the remote server.

So we need to keep these two 'database' consistent, I think you need to refer to [this line](https://github.com/cytoscape/cytoscape-web/blob/development/src/components/ToolBar/DataMenu/LoadWorkspaceDialog.tsx#L107): add a function call here to delete the target workspace in indexed DB
```
if (selectedWorkspace) {
        const ndexClient = new NDEx(ndexBaseUrl)
        const token = await getToken()
        ndexClient.setAuthToken(token)
        await ndexClient.deleteCyWebWorkspace(selectedWorkspace.workspaceId)
        //
        // add a function call here to delete the workspace in indexed DB
        //
      }
```

GAOChengzhan:
Hi Yihang @yihangx,

I now understand the rationale behind the changes you implemented. While this update addresses the immediate issue mentioned in the ticket, the underlying problem is that the local index DB is out of sync. This misalignment leads to additional bugs. For example, if you have a workspace named 'A' in your account and open another called 'B', then delete 'A', you should be able to rename 'B' to 'A' and save it without encountering any name conflicts. However, currently, this scenario triggers a name conflict warning.

I will bring this up on next Monday meeting and discuss with Jing, Dylan and Kei. We should probably add some `deleteWorkspace` method in `db.ts` and resolve this bug completetly.

GAOChengzhan (approved):

PR #404, opened 2025-01-22
Fix continuous mapping form ui issues

d2fong:
- Set default value of the new handle form to be the mean of the max and min values
- Users can drag the end handles along the y axis
- Fix issue where long column names would overflow the new handle form

### DIFF
--- a/src/components/Vizmapper/Forms/MappingForm/ContinuousMappingForm/ContinuousColorMappingForm.tsx
+++ b/src/components/Vizmapper/Forms/MappingForm/ContinuousMappingForm/ContinuousColorMappingForm.tsx
@@ -1021,7 +1021,14 @@ export function ContinuousColorMappingForm(props: {
                   alignItems: 'center',
                 }}
               >
-                <Box sx={{ maxWidth: 80, overflow: 'hidden' }}>
+                <Box
+                  sx={{
+                    maxWidth: 80,
+                    overflow: 'hidden',
+                    textOverflow: 'ellipsis',
+                    whiteSpace: 'nowrap',
+                  }}
+                >
                   {m.attribute}
                 </Box>
                 <ExpandableNumberInput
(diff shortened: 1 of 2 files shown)


GAOChengzhan:
Hi Dylan,

The first two bugs are resolved perfectly. For the third one: column names would overflow the new handle form. Now it turns like this:

image

It will cut the text. I feel maybe could you add some '...' after the string?  For example, set the threshold of the length of the string is 10, if it is larger than 10, then return the first 7 characters and append '...'.

This is in Cytoscape Desktop
image

GAOChengzhan:
@d2fong I mistakenly pushed a commit to this branch. I reset it and pushed it again. Could you help me check whether it is now the same status as your local version?

GAOChengzhan:
@d2fong Sorry I did not notice this earlier:
I think the 'Set Min and Max' dialog also needs to be changed

image

Basically it's
1. this [line](https://github.com/cytoscape/cytoscape-web/blob/CW-489/src/components/Vizmapper/Forms/MappingForm/ContinuousMappingForm/ContinuousColorMappingForm.tsx#L1120) in ContinuousColorMappingForm.tsx
2. this [line](https://github.com/cytoscape/cytoscape-web/blob/CW-489/src/components/Vizmapper/Forms/MappingForm/ContinuousMappingForm/ContinuousNumberMappingForm.tsx#L809) in ContinuousNumberMappingForm.tsx
3. and at the same time, the dialog in ContinuousColorMappingForm.tsx needs to have a width limit [here](https://github.com/cytoscape/cytoscape-web/blob/CW-489/src/components/Vizmapper/Forms/MappingForm/ContinuousMappingForm/ContinuousColorMappingForm.tsx#L1118), otherwise it can be super wide (the dialog in ContinuousNumberMappingForm does have a fixed width of 200)

GAOChengzhan (approved):

PR #417, opened 2025-02-01
Partial fix for CW-513

d2fong:
- Update the default color scheme colors to be the same ones as Cytoscape Desktop
- Set the min and max values of the continous color creation function to be the largest absolute value if the data is two tailed
- Unable to fix the selected palette as the palette UI code needs to be reworked


GAOChengzhan (approved):
Tested and it works well

PR #418, opened 2025-02-01
Fix bug of adding an extra network when loading workspace

GAOChengzhan:
Ticket: [CW-516](https://cytoscape.atlassian.net/browse/CW-516)

This bug can not be reproduced steadily. In my experience, it is easier to reproduce on a local machine (with longer internet latency).

The root of the bug lies in:
When loading a new workspace, due to the change of `workspaceId`, the `redirect()` function in `AppShell.ts` would be triggered. And it will parse the `networkId` in the URL (src/components/AppShell.tsx line 242), and as long as the `currentNetworkId` is valid, it will check whether the parsed network is up-to-date and load it to the workspace.

According to the comment from line 272 to 274 in src/components/AppShell.tsx

>         // the user is trying to load a network that is already in the workspace
>         // check that if they have an outdated version of the network by comparing modification times
>         // of the local copy and the ndex summary

This network should be loaded if it is already in the workspace. But the condition "it is in the workspace" has not been checked, so I added this condition checker: 
```
} else if (networkIds.includes(networkId)) {

```

I am not 100% sure whether this will harm the original usage of this function. Please help me check this.

[CW-516]: https://cytoscape.atlassian.net/browse/CW-516?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/AppShell.tsx
+++ b/src/components/AppShell.tsx
@@ -268,7 +268,7 @@ const AppShell = (): ReactElement => {
         navigate(
           `/${id}/networks/${currentNetworkId}${location.search.toString()}`,
         )
-      } else {
+      } else if (networkIds.includes(networkId)) {
         // the user is trying to load a network that is already in the workspace
         // check that if they have an outdated version of the network by comparing modification times
         // of the local copy and the ndex summary


GAOChengzhan:
----------------------UPDATE----------------------
The reason why the bug cannot be reproduced steadily is that in line 242 in src/components/AppShell.tsx:
```
 const parsedNetworkId = urlNetIdRef.current
```
Sometimes it would be an empty string which would not causing adding an additional network
WHILE sometimes it would fetch the `networkId` in the previous URL which finally results in adding an additional network

keiono (changes requested):
Please drop this pull request and test mine. It should resolve this issue by waiting and redirect BEFORE updating ws.

GAOChengzhan:
> Please drop this pull request and test mine. It should resolve this issue by waiting and redirect BEFORE updating ws.

Okay, I closed this one.

PR #423, opened 2025-02-07
Clear urlNetIdRef after use

GAOChengzhan:
This is a fix for this [ticket](https://cytoscape.atlassian.net/browse/CW-516)

#### 1. Clear urlNetIdRef after use
Set the `urlNetIdRef` value to an empty string after use in the `redirect` function.
This could resolve the bug and also would not break the original "share network logic"

#### 2. Debug 'Load Network'
- Add 'Waitsecond' before setCurrentNetworkId if the networkId is just added into the workspace
- Clear the UpdateNetwork dialog
    - Delete the unnecessary code since the dialog is shown only when the user is authenticated
    - Pass the correct target networkId to the dialog  instead of reading the currentNetworkId (the target networkId is not necessarily the current one)

### Related Pull Request:
[Fix bug of adding an extra network when loading workspace](https://github.com/cytoscape/cytoscape-web/pull/418)
[CW-516 Fixed timing issue after loading workspace](https://github.com/cytoscape/cytoscape-web/pull/420)

### DIFF
--- a/src/components/AppShell.tsx
+++ b/src/components/AppShell.tsx
@@ -58,6 +58,7 @@ const AppShell = (): ReactElement => {
 
   // This is necessary to prevent creating a new workspace on every render
   const [showDialog, setShowDialog] = useState<boolean>(false)
+  const [targetNetworkId, setTargetNetworkId] = useState<string>('')
   const [search, setSearch] = useSearchParams()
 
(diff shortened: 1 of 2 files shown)


keiono (approved):

keiono:
The App Shell's reloading logic must be re-designed in the next round since there is too many uncertainties, but for now, this can be a temp. fix.

PR #426, opened 2025-02-13
Fix issue where annotations may not be rendered

d2fong:
- Depending on how the network summary is retrieved from NDEx, it may give a network attribute value as a single string even when it specifies the data type as an array

Fixes CW-508

### Testing
Load network 495f1d07-eb3b-11eb-b666-0ac135e8bacf from dev1 into cytoscape web.  The annotation should be displayed.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "cytoscape-web",
-  "version": "1.0.0-beta.1",
+  "version": "1.0.1",
   "lockfileVersion": 3,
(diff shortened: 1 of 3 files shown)


GAOChengzhan:
@d2fong Hi Dylan, I switch to your branch and check the network, it does not render the big rounded rectangle, could you check this on your side?
image

GAOChengzhan (approved):
Tested on that network and it can show the big rounded rectangle

PR #428, opened 2025-02-20
Fix issue where users could input floats into integer columns

d2fong:
- Perform a check when the user inputs values into integer columns in the table browser
- Reject the value if it is a float.

Testing:
- Open a network
- Input a decimal value in a integer column
- It should not update the value

### DIFF
--- a/src/components/TableBrowser/TableBrowser.tsx
+++ b/src/components/TableBrowser/TableBrowser.tsx
@@ -455,31 +455,40 @@ export default function TableBrowser(props: {
         return
 
       if (isListType(column.type)) {
-        data = deserializeValueList(column.type, data as string)
-      }
-
-      if (
-        column.type !== ValueTypeName.Integer &&
-        column.type !== ValueTypeName.Long
-      ) {
-        setCellValue(
-          props.currentNetworkId,
-          currentTable === nodeTable ? 'node' : 'edge',
-          `${cxId}`,
-          columnKey,
-          data as ValueType,
-        )
-      } else {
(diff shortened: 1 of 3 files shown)


src/components/TableBrowser/TableBrowser.tsx
```diff
@@ -454,10 +458,7 @@ export default function TableBrowser(props: {
         data = deserializeValueList(column.type, data as string)
       }
 
-      const newDataIsValid = true
-
-      // TODO validate the new data
-      if (newDataIsValid) {
+      if (column.type !== ValueTypeName.Integer) {
```
GAOChengzhan, line 461: Hi Dylan, @d2fong  maybe we should also add `long` type here?
```
if (column.type !== ValueTypeName.Integer && column.type !== ValueTypeName.Long ) {
```

GAOChengzhan: And a few other small bugs I found are:

1. It can still let me input double item(s) into `list_of_int` or `list_of_long` column
image

2. It also let me create new `list_of_int` or `list_of_long` column with default values
image

d2fong: Good catch.  Those issues should be fixed now.

GAOChengzhan:
Hi Dylan, I think there is one bug left:
I can still set 'list_of_double' to 'list_of_integer' or 'list_of_long' type column
image

GAOChengzhan (approved):

PR #431, opened 2025-03-18
Pie ring charts

yihangx:
1. This PR does not implement ring chart. Will do it later.
2. Like NDEx, https://github.com/cytoscape/js4cytoscape/blob/1110fc685e43ce4d33cdcde91eac1106ead7c0e9/packages/cx2js/src/cx_to_js.js#L1633
It only work for NODE_CUSTOMGRAPHICS_1. Need to fix it later.


GAOChengzhan:
@yihangx Hi Yihang, could you drop some networks with pie chart here for me to test? Thanks!

GAOChengzhan (commented):
I encountered this error when trying to switch the 'style' tab after loading the provided test networks (pie_test.cx2 or test.cx2)
image

I guess it's caused by the newly added visual properties do not have `maxVal` in them. This can be fixed if I changed this [line](https://github.com/cytoscape/cytoscape-web/blob/development/src/components/Vizmapper/VisualPropertyRender/Number.tsx#L49): 
From
```
const maxVal = getDefaultVisualStyle()[vpName].maxVal
```
To
```
const maxVal = getDefaultVisualStyle()[vpName]?.maxVal
```

src/models/VisualStyleModel/impl/DefaultVisualStyle.ts
```diff
@@ -368,9 +374,19 @@ export const getDefaultVisualStyle = (): VisualStyle => ({
...
-    tooltip: 'The stacking order of edges on the canvas. Higher values are drawn on top.'
+    tooltip:
+      'The stacking order of edges on the canvas. Higher values are drawn on top.',
+  },
+  pieSize: {
+    group: 'node',
+    name: 'pieSize',
+    displayName: 'Pie Size',
+    type: 'number',
+    defaultValue: '80%',
```
GAOChengzhan, line 385: Since the type is `number`, the default value must also be a number instead of string like `80%`, otherwise it is illegal

GAOChengzhan:
Another problem I encountered is that the pie-chart info is not exported correctly during exportation

Which is to say:

1. Import a network with pie charts to the workspace
2. Download it (Data -> Download Network File (.cx2))
3. Import the just downloaded cx2 file , it does not have the pie-chart info

|Step 1| Step 3|
|-----|-----|
|image | image|

src/models/VisualStyleModel/impl/VisualStyleFnImpl.ts
```diff
@@ -379,11 +377,83 @@ export const createVisualStyleFromCx = (cx: Cx2): VisualStyle => {
...
+    ) {
+      pieChartConfig.cy_colors.forEach((color: string, i: number) => {
+        const vpName = `pie-${i + 1}-background-color`
+        const displayName = `Pie Slice ${i + 1} Background Color`
+        const tooltip = `The background color for pie slice ${i + 1}.`
+        ;(visualStyle as any)[vpName] = {
+          group: 'node',
+          name: vpName,
+          displayName,
+          type: 'string',
```
GAOChengzhan, line 449: This should be type of `color` instead of string, otherwise the user cannot pick the color 
image

GAOChengzhan:
Hi Yihang, I don't know whether you noticed the following thing already or you plan to leave this until all the image/chart -related features are implemented

In Cytoscape Desktop, users have the choice to define up to 9 images/charts. And by clicking any of them, they can further define its detail, like the slice in the case of pie chart. 

Here you listed all the slice features directly in the node style, which is not a preferable pattern.  A better way might be:

- Node Style
    - Chart 1
    - Chart 2
        -  Dialog 
            -  Slice 1
            - Slice 2
            - ...
    - ...


|Cytoscape Web| Cytoscape Desktop|
|-----|-----|
|image| image|

yihangx:
@GAOChengzhan Hide or delete pie chart related properties in visual style UI will break the application. It is not worth doing it I think. Maybe we should discuss how to create a UI for custom graphics in this version.

src/components/NetworkPanel/CyjsRenderer/CyjsRenderer.tsx
```diff
@@ -218,11 +218,47 @@ const CyjsRenderer = ({
       visualEditorProperties,
     )
 
-    // Generate a new Cytoscape.js styles based on given visual style
-    const newStyle = createCyjsDataMapper(vs)
-    setCyStyle(newStyle)
 
-    // Restore selection state in Cyjs instance
+    const newStyle = createCyjsDataMapper(vs);
```
d2fong, line 222: This code should belong in cyjs-util.ts, compute-view-util.ts, DataMapper.ts and DirectMappingSleector.ts

src/models/VisualStyleModel/VisualPropertyName.ts
```diff
@@ -19,6 +19,7 @@ export const NodeVisualPropertyName = {
   NodeMaxLabelWidth: 'nodeMaxLabelWidth',
   NodeZOrder: 'nodeZOrder',
   NodeLabelPosition: 'nodeLabelPosition',
+  PieSize: 'pieSize',
```
d2fong, line 22: This should be a more accurate visual property name that corresponds to its CX visual property name

src/models/VisualStyleModel/impl/DefaultVisualStyle.ts
```diff
@@ -368,9 +374,19 @@ export const getDefaultVisualStyle = (): VisualStyle => ({
     type: 'number',
     defaultValue: 0,
     bypassMap: new Map(),
-    tooltip: 'The stacking order of edges on the canvas. Higher values are drawn on top.'
+    tooltip:
+      'The stacking order of edges on the canvas. Higher values are drawn on top.',
+  },
+  pieSize: {
+    group: 'node',
```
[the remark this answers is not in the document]
d2fong: Define another visual property interface for this type.

See NodeLabelPositionType.ts, VisualPropertyValueType.ts for examples.

src/models/VisualStyleModel/impl/cxVisualPropertyConverter.ts
```diff
@@ -305,6 +305,13 @@ export const VPNodeLabelPositionConverter = (
   }
 }
 
+export const VPPieSizeConverter = (): CXVisualPropertyConverter<string> => {
```
d2fong, line 308: This converter should just return the object you receive from the CX json for the CX vp 'NODE_CUSTOMGRAPHICS_1'

GAOChengzhan:
> TODO decide on the following:
> 
> 1. How will the UI look for pie charts/ring charts?
> 2. Are we going to support images for node custom graphics?
> 3. Of the current options in Cytoscape desktop for custom graphics, which ones will Cytoscape Web support.
> 4. How does Cytoscape Desktop handle passthrough mappings, discrete mappings and continuous mappings for custom graphics?  Does it make sense for Cytoscape Web to handle them in the same way?  If not, how should Cytoscape Web handle them?
> 
> These things must be discussed and decided upon before any code should be merged. As it will have implications for how this feature should be implemented.

Yes, we should discuss before moving on. The code needs to be restructured/redesigned.

PR #436, opened 2025-04-10
Hotfix for issue where the table browser would block the UI

d2fong:
- Manually call 'setNetworkModified' in the various table browser functions instead of listening to all changes in the table.

Should fix CW-542


keiono (approved):
Need to go back to fix the root cause, but for now, this looks OK to resolve the frozen UI issue.

PR #438, opened 2025-04-17
Undo cell edits

d2fong:


GAOChengzhan (approved):
Tested with all different cell data types and it works well!

PR #440, opened 2025-04-22
Add undo support for network summary editing

d2fong:
- Use cancel/confirm pattern for the network summary editing form


GAOChengzhan:
@d2fong Hi Dylan, I find some bugs in this PR:
1. Undo does not work for changes on the network `description` which originally is empty (It does work if the original description is non-empty). You could reproduce this bug by choosing a network whose description is empty.
2. Undo does not work for changes on `Version`.

d2fong:
> @d2fong Hi Dylan, I find some bugs in this PR:
> 
> 1. Undo does not work for changes on the network `description` which originally is empty (It does work if the original description is non-empty). You could reproduce this bug by choosing a network whose description is empty.
> 2. Undo does not work for changes on `Version`.

Ok I believe I fixed it.  The issue is that the ndex client will return undefined values for version/description if they are empty, but the cytoscape web model assumes they are required fields.  I set default empty values when getting summaries from the ndex client.

GAOChengzhan (approved):
Nice, the bug is resolved!

PR #452, opened 2025-05-22
Don't show any custom graphic vps as there is no ui to let the user m…

d2fong:
…eaningfully edit them.


keiono (approved):
Need to be implemented later, but for now, it is hidden and does not break other parts of UI.

PR #455, opened 2025-05-24
refactored code for citation window. 

jingjingbic:
Make the journal name italic. 
Added URL link to pubmed ids of both citations. 
Removed the "ahead of print". 
Use a single button to copy both citations.


jingjingbic:
Dylan we make further changes in the CW-571 branch and create another PR.

PR #461, opened 2025-06-13
Cx2 validator checkpoint

d2fong:
- add CX2 validator with tiered validation phases (structural validation, meta data, referential integrity, attribute value type checking)
- apply validation logic to services and cx2 file upload


src/models/CxModel/Cx2/Validator.ts
```diff
@@ -0,0 +1,38 @@
...
+  /** Whether to perform strict validation */
+  strict?: boolean
+  /** List of supported CX versions */
+  supportedVersions?: string[]
+}
+
+/**
+ * Represents a validation error or warning
+ */
+export interface ValidationIssue {
```
keiono, line 17: For now, it is OK to use task-specific errors, but we need a generalized  error types for all type of errors.

PR #467, opened 2025-06-25
Misc table features

d2fong:


yihangx (commented):
Tested on exporting some networks with columns manipulations and import them back. Looks great!

yihangx (approved):

PR #469, opened 2025-07-01
Sif file import

d2fong:


jingjingbic:
We should have a default style that is similar to the one in Cytoscape Desktop applied to SIF file. This is what I am proposing:

Node shape: rounded rectangle, color light blue (same as Cytoscape), label text: black, Node border: none. Edge color: dark grey. node label passthrough mapping from node name.

yihangx (approved):
Sif file import works.

PR #473, opened 2025-07-09
fix CW-600

yihangx:
https://cytoscape.atlassian.net/browse/CW-600


jingjingbic (commented):
Using the the urlBaseName instead of relying on regular express is a safer and better implementation. You can follow the same pattern used in other components like `LoginButton.tsx`.
The improved implementation would be something like:

```typescript
// Import the context
import { AppConfigContext } from '../../AppConfigContext'

// In the component
const { urlBaseName } = useContext(AppConfigContext)

// In handleClick
const handleClick = (): void => {
  const { location } = window
  const baseUrl = location.origin + urlBaseName
  // ... rest of the code
  const newUrl = `${baseUrl}0/networks/${currentNetworkId}?${finalQuery}`
}
```

PR #507, opened 2026-03-06
fix(FE-005): replace any type with proper typing for routerOpts

kushvinth:
## Change

`src/App.tsx`,  replaced `const routerOpts: any = {}` with `const routerOpts: { basename?: string } = {}`.

## Why

The `routerOpts` object only ever receives a `basename` property. Using `any` bypasses type checking unnecessarily.

## Testing

- No behavioral change, type-only fix.

## Related Issue
- #506

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,0 +1,6 @@
+[build]
+  command = "npm run build:netlify"
+  publish = "dist"
+
+[build.environment]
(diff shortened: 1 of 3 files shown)


kushvinth:
PTAL @coleslaw481, @jingjingbic, @d2fong and other maintainers

PR #513, opened 2026-03-16
fix(CW-643): add selection counts to table browser tabs

d2fong:
Adds counts of selected nodes and edges to the respective tab labels in the Table Browser. Counts are only shown when greater than zero.


d2fong:
@copilot

d2fong:
Screenshot 2026-03-16 at 4 51 03 PM

PR #518, opened 2026-03-31
CW-699: Interpolate color palettes when slices exceed base colors

d2fong:
Matches Cytoscape Desktop behavior by interpolating colors using `chroma-js` when a color palette is smaller than the requested count instead of looping. \n\n @copilot review


d2fong:
@copilot please review

PR #529, opened 2026-04-27
New app api

d2fong:


keiono (approved):
Need to update the example apps, but new API is working now.

PR #531, opened 2026-04-30
Fix/cw 683 - New node is not always created within the viewport

chrtannus:
The new node is now added to top-left corner of the viewport, plus a small padding.
I think this is better than adding the node to the center, because the center of the viewport is usually more crowded than the corners, which means the new node would overlap other nodes more often.

When testing, please make sure to test with different zoom levels (after zooming in and out) and after panning the network. Also test with the option "Lock node width and height" checked and then unchecked--when unchecked, make sure the width and height have different values.

Issue: [CW-683](https://cytoscape.atlassian.net/jira/software/projects/CW/issues?jql=project+%3D+CW+ORDER+BY+updated+DESC%2C+cf%5B10005%5D+ASC&selectedIssue=CW-683&atlOrigin=eyJpIjoiZGRiN2VjZDM4NTZiNGYyZDkxMzc1YWJmMmMzZWIzYWUiLCJwIjoiaiJ9)

[CW-683]: https://cytoscape.atlassian.net/browse/CW-683?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


d2fong:
Moving this PR to 1.0.7 https://github.com/cytoscape/cytoscape-web/pull/532

PR #544, opened 2026-06-06
added ci gh workflows

sreuland:
wire up gh ci workflow on pr's and merge to master to run unit/e2e tests in project. 

The e2e test didn't have any tests yet, created several tests based on recent commits and config for verify on main browsers(chromium, webkit, firefox) on ci gh runner.

Closes https://cytoscape.atlassian.net/browse/CW-748


maxkfranz:
This would be a good PR to set the `.nvmrc` and reference the `.nvmrc` in the actions yml.  That way, the node version is specified in one place for everything: build envs like Cloudflare or Netlify, CI envs like Github Actions, local dev envs via `nvm` or  (esp.) `mise`.

E.g. https://github.com/cytoscape/cytoscape.js/blob/unstable/.github/workflows/tests.yml#L11-L13

maxkfranz:
The timeout duration is also important to set.  Otherwise, hanging tests can eat up all our usage allowance.

E.g. https://github.com/cytoscape/cytoscape.js/blob/unstable/.github/workflows/tests.yml#L8

sreuland:
> This would be a good PR to set the .nvmrc and reference the .nvmrc in the actions yml. That way, the node version is specified in one place for everything: build envs like Cloudflare or Netlify, CI envs like Github Actions, local dev envs via nvm or (esp.) mise.

@maxkfranz , good call, I updated a few more places to drive consistency to same node version as well such as engine level in package.json and updated README for visibility how node version is required.

https://github.com/cytoscape/cytoscape-web/pull/544/commits/6de01e9df4070c2303088ec7ecc19e8f4fe8bdb9

sreuland:
> The timeout duration is also important to set. Otherwise, hanging tests can eat up all our usage allowance.

@maxkfranz , I added timeouts on ci jobs, also, parallelized the integration test to run `test:e2e` for each playwright browser at same time. In terms of usage allowance concerns, do we want ci to maintain e2e browser coverage on all 3 chromium, webkit, and firefox? small change to the matrix in the job to restrict which browsers if so.

maxkfranz:
> In terms of usage allowance concerns, do we want ci to maintain e2e browser coverage on all 3 chromium, webkit, and firefox? small change to the matrix in the job to restrict which browsers if so.

I think chromium alone is fine.  

Playwright tests are really slow, and the additional time/cost isn't worth it.  The most common browser-engine-specific bugs are rendering-related and difficult to automatically test anyway.

Maybe all engines as a separate npm script that could be run only just before a release (maybe part of a github action release yml in future)?

I don't think it's reasonable to have devs or github run all browsers by default via `npm test`.  If it's slow, people generally won't bother.

sreuland:
> Playwright tests are really slow

each browser engine is currently at 3-4 minutes for e2e test runs, and each is in parallel, of course as people start adding more e2e tests this will change

> I don't think it's reasonable to have devs or github run all browsers by default via npm test

there are two test targets `test:unit` which runs js unit tests only, no playwright so that remains same. the `test:e2e` target is basically new here, it has no tests yet, this is first introduction of e2e level coverage with playwright, and this is an optional target that people choose when to run it locally.

On the gh ci, there are two jobs to run test suites, one that runs test:unit and then one that runs test:e2e in parallel for each browser engine, for now I can dial back the ci matrix for browser engines to just chromium and the `test:e2e` will default that to run just chromium for local users. The readme has new section for docs on running e2e tests and that shows how to run test:e2e with the -`--project` param to choose specific browser engines if users want that coverage.

maxkfranz (approved):
LGTM

PR #548, opened 2026-06-11
test: migrate Jest to Vitest

d2fong:
- use Vite test instead of jest
- Runs ~2k unit tests in 30s down from 95s


maxkfranz:
Is it intended to use `@testing-library/jest-dom`?  With vitest, I think you can just `npm i -D jsdom` and vitest will use it if you set it in the config, e.g:

```ts
// a pretty standard config

import { defineConfig, mergeConfig } from 'vitest/config';
import viteConfig from './vite.config';

export default defineConfig(async () => {
  const resolvedViteConfig = await (typeof viteConfig === 'function'
    ? viteConfig({ command: 'serve', mode: 'test' })
    : viteConfig);

  return mergeConfig(resolvedViteConfig, {
    test: {
      environment: 'jsdom',
      include: ['test/**/*.test.{ts,tsx}'],
      setupFiles: ['./test/setup.ts']
    },
  });
});

```

https://vitest.dev/guide/environment

Issue #550, opened 2026-06-16
discrete mapping on node fill color is rendered as pie chart.


jingjingbic:
I imported a cx2 network which has a discrete mapping on node fill-color. In Cytoscape Web the nodes are rendered as piechart. It is reproducable in 1.0.7 and 1.0.8 snapshot.

jingjingbic:
[6e36f9fe328c402fa486b01524412f61.graphml.cx2.zip](https://github.com/user-attachments/files/29020346/6e36f9fe328c402fa486b01524412f61.graphml.cx2.zip) cx2 file attached.

jingjingbic:
Root cause: the attached cx2 file has a pie chart definition with empty data column list. Cytoscape Desktop ignores it when rendering it. We should do something similar in Cy Web.

PR #565, opened 2026-07-01
Fix: Skip pie/ring chart rendering when cy_dataColumns is empty or un…

d2fong:
…defined

refs #550


jingjingbic:
@d2fong , could you add the unit test according to Copilot's comments?

jingjingbic (changes requested):
@d2fong , could you add unit test for the changes?

PR #568, opened 2026-07-02
Adding acceptance test using the playwright framework

jingjingbic:
Adding 2 lanes of testing. 1. Basic UI test using playwright. 2. Optional acceptance test using playwright and NDEx dev1 server.


sreuland:
if you want the `test:e2e:acceptance` to be runnable as ad-hoc requests, check out .github/workflows/ci.yml, could clone that to acceptance.yml on this pr and use the existing `integration-tests` job in there that runs playwright package.json target also. then add `on:dispatch` in that .yml which enables running the job as a 'Run Workflow' button on the GH Actions tab for cytoscape-web repo. Here's an example of [workflow dispatch definition](https://github.com/cytoscape/platform-automation/blob/main/.github/workflows/jira-port.yml#L12)

d2fong:
Max added some playwright utilities that make them more reliable.  New tests should use the patterns found in the latest commits from development branch.

PR #574, opened 2026-07-13
docs,build: make Windows support automatic; drop stale manual setup

maxkfranz:
The README's "Windows-Specific Setup Instructions" predated the Vite migration and no longer applied. It told Windows users to rewrite their package.json scripts to invoke `webpack` and to manually populate `REACT_APP_GIT_COMMIT` / `REACT_APP_BUILD_TIMESTAMP` in a `.env` file. None of that is true anymore:

  - The project builds with Vite, not webpack.
  - Build metadata is computed at build time inside vite.config.ts via `execSync('git ...')` and injected through `define`. This runs identically on Windows, macOS, and Linux, so no env vars need to be set by hand.
  - The scripts that do set environment variables (build:bundle, build:netlify, build:analyze) already use `cross-env`, and the rest of the toolchain (npm-run-all, rimraf, cpy-cli) is cross-platform.

Following the stale steps would actively break a Windows setup, so the section is replaced with a short note stating that no Windows-specific setup is required and why.

While confirming the daily dev/build/test flow is cross-platform, two scripts still used bash-only single-quote idioms that cmd.exe and PowerShell do not strip:

  - `format`: single-quoted prettier glob
  - `verify:federation`: single-quoted --compilerOptions JSON

Both are switched to double quotes with backslash-escaped inner quotes, the form node's argv parser accepts on Windows and that bash unescapes identically. Verified the format glob still resolves (303 files matched) and that package.json remains valid JSON.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,51 +115,9 @@ npm run test:e2e -- --project=chromium --project=webkit
 
 Available project names are `chromium`, `firefox`, and `webkit` (matching `playwright.config.ts`).
 
-### Windows-Specific Setup Instructions
+### Windows
 
-For Windows users, environment variables need to be set differently. Follow these steps to run the development server.
-
-1.  **Modify `package.json` scripts**
-    Update the `dev` and `buid` scripts in `package.json` like this
-
-    ```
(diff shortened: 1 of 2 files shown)


maxkfranz:
@chrtannus, @d2fong -- would one of you try this out?  If you don't have windows handy, just verify it doesn't break on mac/linux/whatever you have.  Then we can merge and @keiono can carry the windows torch from there, if that's what he uses.

PR #613, opened 2026-07-19
fix(auth): disable keycloak silent SSO full-page redirect fallback (CW-663

d2fong:
With third-party cookies blocked (e.g. incognito), keycloak-js falls back from silent check-sso to a full-page redirect whose redirect_uri is location.href — including user-supplied query params such as ?import=<url>. NDEx Keycloak can reject that redirect_uri with "Invalid parameter: redirect_uri", stranding the user on the Keycloak error page before the app loads.

Setting silentCheckSsoFallback: false keeps the check in the hidden iframe, which resolves unauthenticated without any top-level navigation. Init options are extracted to buildKeycloakInitOptions() with unit tests covering the local-dev and deployed paths.

## What was wrong
The ?import= parameter was leaking into the OAuth redirect_uri. The sequence, confirmed by inspecting the bundle deployed at web.cytoscape.org:

On startup, keycloak.init runs a silent SSO check (check-sso + silent-check-sso.html in a hidden iframe). This requires third-party cookies.
In incognito, third-party cookies are blocked. keycloak-js detects this, and because silentCheckSsoFallback defaults to true (we never set it), it discards the silent iframe approach and performs a full-page redirect to NDEx Keycloak instead.
That fallback uses location.href as the redirect_uri — i.e. https://web.cytoscape.org/?import=http://localhost:8000/valid.cx2, with the user-supplied import URL embedded.
If Keycloak rejects that redirect_uri, the user is stranded on Keycloak's "Invalid parameter: redirect_uri" page — visibly, because this is a top-level navigation, unlike the hidden iframe.
Without ?import=, the redirect_uri is the clean origin, so the bounce succeeds invisibly — which is why plain incognito loads "worked" (they were actually doing a hidden round-trip to Keycloak on every load). The auth server never fetches or cares about the import URL itself; it just fails to validate the page URL carrying it.

## How to reproduce
NDEx production Keycloak currently accepts the ticket's exact URL (verified by direct probes), so that literal URL no longer errors — the server config has likely been loosened since the report, or Manju's URL carried an extra character. Two deterministic reproductions still exist:

Invalid characters (production): Keycloak rejects redirect_uris that fail strict URI parsing, and browsers pass { } | ^ \ and stray % through raw in query strings. In incognito, load https://web.cytoscape.org/?import=http://localhost:8000/valid{1}.cx2 → Keycloak error page. The file doesn't need to exist — the failure happens before any fetch. The same URL in a normal window loads fine.
Branch deploy (clean URL): Netlify branch deploys use dev1 Keycloak, which rejects the Netlify origin outright. In incognito, https://development--incredible-meringue-aa83b1.netlify.app/?import=http://localhost:8000/valid.cx2 shows the identical error with the unmodified URL.
To observe the mechanism on any URL: DevTools → Network with Preserve log, load in incognito, and look for a top-level request to .../auth2/realms/ndex/protocol/openid-connect/auth whose redirect_uri contains the ?import= parameter.

The fix (commit 37af04fc)
Set silentCheckSsoFallback: false in the Keycloak init options. With the fallback disabled, the silent check still runs in the hidden iframe when third-party cookies are blocked and simply resolves unauthenticated — no top-level redirect, no user-controlled redirect_uri, no error page. Incognito users just aren't auto-signed-in (which was already the effective outcome); the explicit Login button flow is unchanged, as is local dev.

Init options were extracted into buildKeycloakInitOptions() in src/init/keycloak.ts with unit tests (src/init/keycloak.test.ts), including a regression test pinning the fallback off. Full unit suite passes (2116 tests), lint and tsc clean.

Verification after deploy: repeat either repro against the fix branch's deploy (cw-663-fix-incognito-import-redirect--incredible-meringue-aa83b1.netlify.app once pushed) — the Network log should show no top-level /auth2/... navigation, and the app should load logged-out with no error.

### DIFF
--- a/src/init/keycloak.ts
+++ b/src/init/keycloak.ts
@@ -1,11 +1,33 @@
 import { NDExAuthError, NDExClient } from '@js4cytoscape/ndex-client'
-import Keycloak from 'keycloak-js'
+import Keycloak, { KeycloakInitOptions } from 'keycloak-js'
 import { createContext } from 'react'
 
 import appConfig from '../assets/config.json'
 import { getNDExBaseUrl } from '../data/external-api/ndex/config'
 
 export const KeycloakContext = createContext<Keycloak>(new Keycloak())
+
+export const buildKeycloakInitOptions = (
+  isLocalDevHost: boolean,
+  urlBaseName: string,
+): KeycloakInitOptions => {
+  if (isLocalDevHost) {
+    return { checkLoginIframe: false }
+  }
+  return {
+    onLoad: 'check-sso',
+    checkLoginIframe: false,
+    // CW-663: with third-party cookies blocked (e.g. incognito), keycloak-js
+    // would otherwise fall back to a full-page redirect check-sso whose
+    // redirect_uri is location.href — including user-supplied params such as
+    // ?import=<url> — which NDEx Keycloak can reject with
+    // "Invalid parameter: redirect_uri". With the fallback disabled, the
+    // silent iframe check still runs and simply resolves unauthenticated.
+    silentCheckSsoFallback: false,
+    silentCheckSsoRedirectUri:
+      window.location.origin + urlBaseName + 'silent-check-sso.html',
+  }
+}
 
 export const initializeKeycloak = () => {
   const { keycloakConfig, urlBaseName } = appConfig
--- a/src/init.tsx
+++ b/src/init.tsx
@@ -14,7 +14,11 @@ import { EmailVerificationModal } from './features/EmailVerification'
 import ErrorBoundary from './features/ErrorBoundary'
 import { FeatureAvailabilityProvider } from './features/FeatureAvailability'
 import { initializeGoogleAnalytics } from './init/googleAnalytics'
-import { initializeKeycloak, KeycloakContext } from './init/keycloak'
+import {
+  buildKeycloakInitOptions,
+  initializeKeycloak,
+  KeycloakContext,
+} from './init/keycloak'
 import { BootScreen } from './init/BootScreen'
 import { initializeTabManager } from './init/tabManager'
 
@@ -121,16 +125,10 @@ const initializeApp = () => {
         renderAppOnce({ authenticated: false })
       }, AUTH_INIT_TIMEOUT_MS)
 
-  const keycloakInitOptions = isLocalDevHost
-    ? {
-        checkLoginIframe: false,
-      }
-    : {
-        onLoad: 'check-sso' as const,
-        checkLoginIframe: false,
-        silentCheckSsoRedirectUri:
-          window.location.origin + urlBaseName + 'silent-check-sso.html',
-      }
+  const keycloakInitOptions = buildKeycloakInitOptions(
+    isLocalDevHost,
+    urlBaseName,
+  )
 
   keycloak
     .init(keycloakInitOptions)
--- a/src/init/init_docs/init.md
+++ b/src/init/init_docs/init.md
@@ -75,8 +75,9 @@ Handles Keycloak authentication setup and user verification.
 **Design Decisions:**
 
 1. **Silent SSO Check**: Uses `check-sso` mode to check authentication without redirect
-2. **Error Message Parsing**: Extracts user info from NDEx error messages to populate verification modal
-3. **Verification Check**: Only checks verification for authenticated users to avoid unnecessary API calls
+2. **No Full-Page Fallback (CW-663)**: `silentCheckSsoFallback` is disabled. When third-party cookies are blocked (e.g. incognito), keycloak-js would otherwise fall back to a full-page redirect whose `redirect_uri` is `location.href` — including user-supplied query params such as `?import=<url>` — which NDEx Keycloak can reject with "Invalid parameter: redirect_uri". With the fallback off, the silent iframe check simply resolves unauthenticated.
+3. **Error Message Parsing**: Extracts user info from NDEx error messages to populate verification modal
+4. **Verification Check**: Only checks verification for authenticated users to avoid unnecessary API calls
 
 ### loadingScreen.ts
 
--- a/src/init/keycloak.test.ts
+++ b/src/init/keycloak.test.ts
@@ -1,0 +1,32 @@
+import { describe, expect, it } from 'vitest'
+
+import { buildKeycloakInitOptions } from './keycloak'
+
+describe('buildKeycloakInitOptions', () => {
+  it('skips SSO check entirely on local dev hosts', () => {
+    const options = buildKeycloakInitOptions(true, '/')
+
+    expect(options).toEqual({ checkLoginIframe: false })
+  })
+
+  it('uses silent check-sso against silent-check-sso.html on deployed hosts', () => {
+    const options = buildKeycloakInitOptions(false, '/')
+
+    expect(options.onLoad).toEqual('check-sso')
+    expect(options.checkLoginIframe).toEqual(false)
+    expect(options.silentCheckSsoRedirectUri).toEqual(
+      window.location.origin + '/silent-check-sso.html',
+    )
+  })
+
+  // CW-663: when third-party cookies are blocked (incognito), keycloak-js
+  // falls back from silent check-sso to a full-page redirect that sends
+  // location.href — including user-supplied params such as ?import=<url> —
+  // as redirect_uri, which NDEx Keycloak can reject with
+  // "Invalid parameter: redirect_uri". The fallback must stay disabled.
+  it('disables the full-page redirect fallback for silent check-sso', () => {
+    const options = buildKeycloakInitOptions(false, '/')
+
+    expect(options.silentCheckSsoFallback).toEqual(false)
+  })
+})


src/init/keycloak.ts
```diff
@@ -1,12 +1,34 @@
...
+  return {
+    onLoad: 'check-sso',
+    checkLoginIframe: false,
+    // CW-663: with third-party cookies blocked (e.g. incognito), keycloak-js
+    // would otherwise fall back to a full-page redirect check-sso whose
+    // redirect_uri is location.href — including user-supplied params such as
+    // ?import=<url> — which NDEx Keycloak can reject with
+    // "Invalid parameter: redirect_uri". With the fallback disabled, the
+    // silent iframe check still runs and simply resolves unauthenticated.
+    silentCheckSsoFallback: false,
```
jingjingbic, line 26: Changing this to false might change the app's behaviour when a user is signed in and hit the refresh button. I think we might need to consider another design for this issue.

d2fong:
Option 2 separates two things that are currently mixed together:
1. Keycloak always redirects to a small, fixed set of trusted callback URLs.
2. Cytoscape Web temporarily remembers which application route the user was actually visiting.
No Keycloak tokens are stored. Only the relative Cytoscape Web route is stored temporarily.
Overall browser flow
Assume the user opens:
https://cytoscape.example/0/networks/abc
  ?left=open
  &selectednodes=131
  &import=https%3A%2F%2Fdata.example%2Fnetwork.cx2Step 1: Capture the application location
Before calling keycloak.init(), Cytoscape Web canonicalizes and stores:
/0/networks/abc?left=open&selectednodes=131&import=...in sessionStorage:
{
  "relativeUrl": "/0/networks/abc?left=open&selectednodes=131&import=...",
  "createdAt": 1784725200000
}Only the relative URL is stored—never the origin, Keycloak tokens, or an external redirect target.
Step 2: Start Keycloak with fixed callbacks
Cytoscape Web initializes Keycloak with:
keycloak.init({
  onLoad: 'check-sso',
  checkLoginIframe: false,
  redirectUri: 'https://cytoscape.example/',
  silentCheckSsoRedirectUri:
    'https://cytoscape.example/silent-check-sso.html',
  silentCheckSsoFallback: true,
})Step 3A: Silent SSO succeeds
When the browser allows Keycloak cookies inside the hidden iframe:
Cytoscape Web route remains unchanged
  → Hidden iframe contacts Keycloak
  → Existing Keycloak session is found
  → New tokens are issued
  → keycloak.init() resolves trueThe saved route equals the current route, so restoring it is effectively a no-op. The storage entry is removed, and the app renders.
Step 3B: Silent SSO is blocked
When the browser blocks third-party cookies:
Cytoscape Web
  → Full-page redirect to Keycloak
  → Keycloak reads its first-party SSO cookie
  → Keycloak redirects to https://cytoscape.example/
  → Cytoscape Web starts againThe callback startup does not overwrite the existing saved route.
keycloak.init() processes the callback, exchanges the authorization code for tokens, and resolves as authenticated.
Step 4: Restore the application route
Before constructing React Router, Cytoscape Web restores:
/0/networks/abc?left=open&selectednodes=131&import=...using:
window.history.replaceState(
  window.history.state,
  '',
  savedLocation.relativeUrl,
)It then removes the sessionStorage entry.
Step 5: Construct the router and render
Only after restoration:
Restore URL
  → Create React Router
  → Render App
  → AppShell reads parameters
  → Import network and restore UI state
  → AppShell removes consumed parametersNew-tab behavior
Suppose the user is already signed into Keycloak and opens a new Cytoscape Web tab.
New tab has no in-memory tokens
  → New tab saves its initial Cytoscape Web route
  → check-sso contacts Keycloak
  → Existing Keycloak SSO cookie is found
  → New tokens are issued
  → Saved route is restored
  → User appears signed insessionStorage is tab-specific, which is useful here. Each new tab remembers its own Cytoscape Web route.
Refresh behavior
Suppose the user refreshes:
/workspace-1/networks/network-2?left=openThe refresh destroys in-memory Keycloak tokens, but the Keycloak SSO cookie remains.
Refresh
  → Save /workspace-1/networks/network-2?left=open
  → Run check-sso
  → Recognize existing Keycloak session
  → Restore the route
  → Render authenticatedEven if Keycloak requires a full-page fallback, the user returns to the same route.
Code changes
1. Add an authentication return-location utility
Suggested file:
src/init/authReturnLocation.tsIt should provide something like:
export interface AuthReturnLocation {
  relativeUrl: string
  createdAt: number
}

export const saveAuthReturnLocation = (
  urlBaseName: string,
): void => {
  // Canonicalize current query parameters.
  // Validate the path is inside the application.
  // Save pathname + search in sessionStorage.
}

export const restoreAuthReturnLocation = (
  urlBaseName: string,
): void => {
  // Read and validate the record.
  // Reject stale or unsafe values.
  // Call history.replaceState().
  // Remove the record.
}

export const clearAuthReturnLocation = (): void => {
  // Remove the sessionStorage record.
}Use a namespaced key:
const AUTH_RETURN_LOCATION_KEY =
  'cyweb:keycloak:return-location'Validation rules:

The value must begin with /.
It must be underneath urlBaseName.
It must not contain a protocol or hostname.
It should expire after approximately five minutes.
OAuth callback fragments must not be stored.
It must be removed immediately after restoration.
2. Canonicalize search parameters
Before saving the route:
const url = new URL(window.location.href)
url.search = url.searchParams.toString()This safely changes:
import=[https://example.org/{network}.cx2](https://example.org/%7Bnetwork%7D.cx2)into:
import=https%3A%2F%2Fexample.org%2F%7Bnetwork%7D.cx2Then update the address without navigating:
window.history.replaceState(
  window.history.state,
  '',
  url,
)Link producers should also use:
const appUrl = new URL(appBaseUrl)
appUrl.searchParams.append('import', cx2Url)No manual decodeURIComponent() is needed. React Router’s URLSearchParams decodes the value.
3. Build fixed callback URLs
Create a helper that handles both / and /cytoscape/ deployments:
const normalizedBaseName = urlBaseName.endsWith('/')
  ? urlBaseName
  : `${urlBaseName}/`

const appBaseUrl = new URL(
  normalizedBaseName || '/',
  window.location.origin,
)

const silentSsoUrl = new URL(
  'silent-check-sso.html',
  appBaseUrl,
)Then initialize:
keycloak.init({
  onLoad: 'check-sso',
  checkLoginIframe: false,
  redirectUri: appBaseUrl.href,
  silentCheckSsoRedirectUri: silentSsoUrl.href,
  silentCheckSsoFallback: true,
})This change belongs around [src/init.tsx (line 54)](/Users/jingchen/git/cytoscape-web/src/init.tsx:54).
4. Save before initialization
The initialization order becomes:
const returnLocation = saveAuthReturnLocation(urlBaseName)

const { keycloak, ...handlers } = initializeKeycloak()

keycloak.init({
  // Fixed options
})On the Keycloak callback reload, an unexpired pending record already exists. Do not overwrite it with the fixed callback URL.
5. Restore before constructing React Router
After keycloak.init() resolves:
.then(async (authenticated) => {
  restoreAuthReturnLocation(urlBaseName)

  // Only now construct and render the application.
})Currently, [App.tsx (line 35)](/Users/jingchen/git/cytoscape-web/src/App.tsx:35) creates the router at module initialization:
const router = createBrowserRouter(...)That needs to be delayed.
One clean design is:
export const createAppRouter = (): Router =>
  createBrowserRouter(...)Then, after URL restoration:
const router = createAppRouter()

root.render(
  <App router={router} />,
)Alternatively, App can lazily create the router when it first renders. The important rule is that createBrowserRouter() must run after the URL is restored.
6. Update manual login
The login button currently invokes client.login() at [LoginButton.tsx (line 31)](/Users/jingchen/git/cytoscape-web/src/features/Login/LoginButton.tsx:31).
Change its flow to:
saveAuthReturnLocation(urlBaseName)

await client.login({
  redirectUri: appBaseUrl.href,
})That returns a user to the same network after manually clicking Login.
7. Keep AppShell parameter processing unchanged
AppShell already processes:

import
left, right, bottom
selectednodes, selectededges
selectedSubnetworkNodes, selectedSubnetworkEdges
activeTableBrowserTab
activeNetworkViewTab
activeNetworkView
filterFor, filterBy, filterRange
Because the complete query string is restored before the router starts, no authentication-specific code needs to know what these parameters mean.
Keycloak configuration changes
For an application at the origin root, register:
Valid Redirect URIs:
https://cytoscape.example/
https://cytoscape.example/silent-check-sso.html

Web Origins:
https://cytoscape.example/For a subpath deployment:
Valid Redirect URIs:
https://cytoscape.example/cytoscape/
https://cytoscape.example/cytoscape/silent-check-sso.html

Web Origins:
https://cytoscape.example/You no longer need the dynamic route or query string to be accepted as a Keycloak redirect URI.
Suggested verification cases
Test at least:
1. Already signed in → open a new tab.
2. Already signed in → refresh a network route.
3. Already signed in → open an encoded import URL.
4. Browser allows silent SSO.
5. Browser blocks third-party cookies and requires full-page fallback.
6. Not signed in → app remains anonymous on the original route.
7. Click Login from a network route and return to that route.
8. Multiple parameters survive together:
?left=open&selectednodes=131+132&activeTableBrowserTab=1&import=...9. Multiple import parameters remain intact.
10. Stale or malicious sessionStorage return values are rejected.
The central principle is:
Keycloak receives only stable callback URLs.
Cytoscape Web owns and restores Cytoscape Web navigation st

Issue #665, opened 2026-08-05
Local file import fails to display the network on first load (races the debounced IndexedDB persist)


maxkfranz:
## Summary

Importing a local network file (Data ▸ Import ▸ Network from File...) succeeds — the summary row and table data appear — but the network view shows:

> Failed to load network data: Local network "…" (…) is not found in cache. Local networks are stored only in your browser and cannot be retrieved from NDEx. If you cleared your browser data, this network may have been lost.

A page reload fully recovers: the network then renders normally.

## Reproduction

Production build (`npm run build` + `vite preview`), fresh IndexedDB:

1. Data ▸ Import ▸ Network from File...
2. Select `test/fixtures/cx2/valid/gal-filtered-chart.valid.cx2`
3. The workspace row appears (N/E counts correct, node table populates), but the view area shows the "Failed to load network data … not found in cache" message.
4. Reload the page → the network renders correctly.

Reproduced 3/3 on `perf/load-and-bundles` **and 3/3 on the pre-#655 baseline (`43f28f0e`)** via a scripted Playwright run, so this predates the #603 performance series.

## Root cause (analysis)

`FileUpload.handleCX2File` adds the network to the in-memory stores (`addNewNetwork(...)` etc.) and then navigates. The URL change triggers `WorkspaceEditor`'s `swapCurrentNetworkHook`, whose `loadCurrentNetworkById` **always reads the network from IndexedDB** (`useLoadCyNetwork` → `getCyNetworkFromDb`) — there is no in-memory short-circuit. The NetworkStore persist is debounced, so the first load races the DB write and loses; the cache-miss path then classifies the network as a lost local network. On reload the write has long since landed, hence the recovery.

## Why the e2e suite doesn't catch it

`test/playwright/local-file-import.spec.ts` asserts the workspace network count and the summary name, but never that the network view actually rendered. A `canvas`/view assertion after import would have caught this.

## Suggested fix directions

- Have `loadCurrentNetworkById` consult the in-memory `NetworkStore` first and fall back to the DB read (the load path already tolerates either source), or
- flush/await the pending persist before `navigateToNetwork` in the import handlers.

Found during manual QA for #603.

Issue #603, opened 2026-07-16
UX: bundle size, startup, initial load, usability


d2fong:
## Purpose

What a user experiences before they can do anything useful — download, startup, first render, first successful task — has never been measured or reviewed here.

## Context

**Bundle**

- `npm run build:analyze` already works: `ANALYZE=true vite build` runs `rollup-plugin-visualizer` (treemap, gzip + brotli) into `ba/bundle-report.html` (gitignored). Nobody has recorded the numbers.
- Manual vendor chunk-splitting was **deliberately removed** in the Vite 8 / Module Federation migration (`vite.config.ts:142-145`) — the federation plugin disables `manualChunks`; grouping async-init shared modules caused circular-async hangs. Chunking is whatever the MF plugin + Rolldown decide, so any splitting idea has to work within that constraint.
- Lazy loading exists but is coarse: `AppShell`/`WorkspaceEditor` (`src/App.tsx`), `NetworkPanel`/`TableBrowser`/two dialog forms (`WorkspaceEditor.tsx:39-61`), image-export forms, federated remote components. Everything else — all of Vizmapper, HierarchyViewer, MergeNetworks, the tiptap editor stack — loads eagerly with the workspace editor.
- The dependency list is heavy and overlapping: at least four UI systems ship together — MUI (+icons), Mantine (core/hooks/modals/notifications/dropzone/tiptap), PrimeReact (+primeicons), glide-data-grid — plus the full tiptap stack, visx, react-color, three graph/layout engines (cytoscape + plugins, @antv/g6, @cosmograph/cosmos), keycloak-js, openai, lodash, and d3. Nobody has quantified what each costs or mapped where each is actually used. (Heads-up: dependency changes need team buy-in, so consolidation would start as analysis regardless.)

**Startup path** (what stands between page-load and interactive)

1. `src/index.tsx` → dynamic `import('./init')`.
2. `init.tsx`: `window.CyWebApi` assignment, BootScreen render, immer/debug/tab-manager/GA init, then **Keycloak `check-sso` with a silent-SSO iframe — with a 4000ms timeout** (`AUTH_INIT_TIMEOUT_MS`, `init.tsx:114-122`) before rendering the app anyway. Authenticated users additionally pay a `checkUserVerification()` network round trip before first render.
[...]

PR #666, opened 2026-08-05
perf: faster initial load and a 25% smaller bundle (#603)

maxkfranz:
## Executive summary

Re #603. Nine commits that make the app load faster and ship less code, each independently measured and verified.

| Metric (prod build, 4 Mbps / 100 ms, median of 5 cold loads) | Before | After |
|---|---|---|
| First render (`react-render`) | 1650 ms | **1374 ms** (−17%) |
| Workspace editor interactive | 4248 ms | **3518 ms** (−17%) |
| Cold-load JS transfer | 1292 KB | **929 KB** (−28%) |
| Total shipped JS (gzip, all chunks) | 2294 KB | **1724 KB** (−25%) |
| UI systems | 4 (MUI, Mantine, PrimeReact, glide) | **2** (MUI, glide) |
| npm packages removed | — | **~100** |

### What changed

**Code splitting & boot path**
- Fixed three defeated lazy boundaries (cytoscape leaked past the NetworkPanel boundary via `RendererStore → DefaultRenderer`; two image-export boundaries bypassed by static imports) and thinned the boot-critical DB import graph so cytoscape, the snapshot module, and the models barrels are off the pre-render path.
- Lazy-loaded Vizmapper, MergeDialog, HierarchyViewer, LLM query panel, and all upload/import dialogs behind their tab/first-open gates (with idle prefetch for the STYLE tab).
- Boot reordering: auth starts in parallel with the DB open, Google Analytics is deferred to post-render idle, and the current network's IndexedDB read starts at boot PUBLISH instead of after editor mount.
- Gated three dialogs that were mounted (and therefore fetched) on every cold load despite never being open.

**Dependency consolidation**
- `@antv/g6` → `@antv/layout`: the full G6 rendering stack (381 KB gz) was bundled to run three layout algorithms headlessly; now a 31 KB chunk. This chunk loads on the default layout path for networks under 1000 elements, so it is close to a cold-path win in practice.
- Removed **Mantine** (upload dialogs, table-import wizards, and the rich-text editor chrome rebuilt on MUI; tiptap itself stays) and **PrimeReact** (toolbar menus rebuilt on MUI with proper `menu`/`menuitem` ARIA roles — the old menus only had them implicitly). Removed 8 further unused packages and declared the previously transitive `d3-interpolate`.

**Bugs found and fixed along the way**
- `NetworkPropertyEditor` silently wiped in-progress user edits whenever a background summary update (e.g. layout completion) landed while the editor was open; the draft now syncs only on open.
- A documented negative result: `modulepreload`/`prefetch` hints for the chunk waterfall were measured and rejected (bandwidth-bound, both variants regressed) — recorded in `STARTUP_SPECIFICATION.md` so it isn't retried blind.

### Verification

Every commit passed lint, the full unit suite (3449 tests, including new `@antv/layout` fidelity tests), `verify:federation`, and targeted Playwright chromium suites for the touched areas. A final manual Playwright walkthrough with screenshot inspection covered menus (light + dark), submenus, upload dialogs, the table-import wizard, the rich-text editor, Vizmapper, layout runs, and network rendering.

Known issues, both confirmed pre-existing on the pre-series baseline: the `hierarchy-cell-view` and `remote-app-load` spec failures, and #665 (local file import races the debounced persist on first load — filed from this work's manual QA).

`STARTUP_SPECIFICATION.md` reference measurements are re-recorded; `boot.md` sequence updated.

https://claude.ai/code/session_01SWcM9jzb7yXJEnANJp43qR

## Summary by CodeRabbit

* **New Features**
  * Added drag-and-drop and browse-based file upload dialogs with validation and clear error feedback.
  * Added rich-text editing for network descriptions, including formatting, lists, links, and alignment.
  * Improved table import previews with editable headers, validation indicators, and advanced parsing settings.
  * Added expanded AntV layout options with improved handling for empty or invalid networks.
* **Performance**
  * Faster startup through parallel authentication and database loading.
  * Deferred analytics, lazy-loaded panels, and background prefetching improve responsiveness.
* **UI Improvements**
  * Updated menus, dialogs, forms, and controls with a consistent interface.
  * Added keyboard-accessible menus and improved disabled-control feedback.


maxkfranz:
Addressed the CodeRabbit review in 39e14d5c. Disposition of each finding:

**Fixed**
- `TableUpload` / `FileUpload`: the 5 MB limit the hint promises is now enforced in the validators (`file-too-large` rejection before the file is read into memory).
- `ValueTypeNameForm`: the `space`/`tab` list-delimiter suggestions now store `DelimiterType.Space`/`DelimiterType.Tab` instead of the literal labels (display still shows the words). This was a pre-existing bug carried through from the Mantine-era Autocomplete — `parseValue()` splits on the stored string, so list imports using those delimiters were corrupted.
- `DropdownMenu`: menu rows and `DropdownMenuItem` are now keyboard-focusable with Enter/Space activation and `aria-disabled`; three new unit tests cover activation, disabled items, and submenu opening.
- `previewTableParts`: the column-header editor is a focusable `role="button"` with Enter/Space, and "Skip first N lines" is clamped to ≥ 0 (a negative value made `slice()` take rows from the file's end).
- `PngExportForm`: clearing a number field no longer collapses size/zoom to 0, and the inches fields step by 0.01.
- `NodeLabelPosition` / `ExpandableNumberInput`: accessible names added.
- `TableColumnAppendForm`: join-flow strings corrected ("key column", "Joining table data…") along with the "it's corrensponding" typos here and in `TableColumnAssignmentForm`.
- Stale Mantine comments updated (`ImportDatabaseMenuItem`, `DataMenu`, `WorkspaceEditor`), and the model imports this PR touched now use the `@/` alias.

**Deferred (follow-ups, not in this perf PR)**
- Full arrow-key roving navigation for the menus: activation, focus, and Escape now work from the keyboard; a proper roving-tabindex implementation with arrow-key submenu traversal is a self-contained a11y task worth its own PR.
- A validated normalizer boundary for `Papa.parse` rows (`ParsedRow`): the current typing intentionally replicates the removed primereact `DataTableValue` (also `Record<string, any>`), so this PR introduces no behavior change there. Agreed it's worth doing; it touches the parse/validation pipeline in both wizards and belongs in a focused change.

Also of note: the `TableColumnAppendForm` loading alert never renders because `handleConfirm` toggles `loading` synchronously — string corrected here, but making the join genuinely async is pre-existing behavior left as-is.

Verified: lint, 3452 unit tests, Playwright chromium 23/23 (toolbar-menu, local-file-import, network-property-menu, llm-query-options, smoke).

maxkfranz:
Nitpicks addressed in a3433989:

**Applied (5)**
- `@/` alias for the remaining src-rooted imports this PR introduced (~20 sites across boot, db, prefetch, table-loader, toolbar, and workspace panels), plus the "occured" typo.
- Selected-state background in the three column-editor button groups now uses `action.selected` instead of the hardcoded `#D6D6D6` (a dark-mode legibility fix — the gray was carried over from the Mantine code), with the duplicated `compactButtonSx` moved to a shared module.
- `TableLoaderDialogShell`: `aria-label` on both fullscreen toggles and `aria-labelledby` wiring the Dialog to its title.
- `ParsedRow` is the single canonical row type: the transitional `ParsedRow as DataTableValue` aliases are renamed away and `previewTableParts` re-exports the model type instead of redeclaring it.
- From the outside-diff comments: `aria-label` on the join form's key-column Select, and both wizards now guard the zero-row `Papa.parse` result before the headerless branch dereferences `result.data[0]`.

**Skipped (3, with reasons)**
- Prettier quotes in `DropdownMenu.spec.tsx`: false positive — the only double quotes are JSX attributes, which is the configured Prettier style; the string literals are already single-quoted.
- Papa.parse validation before constructing `ParsedRow`: stays bundled with the already-deferred normalizer follow-up (same pipeline, one focused change).
- The unwired "Case sensitive key values" checkbox: pre-existing (the Mantine-era form had the same decorative checkbox) and a behavior change to the join logic — filed as #667.

Verified: lint, 3452 unit tests, Playwright chromium 13/13 (local-file-import, toolbar-menu, table-browser, smoke).

PR #668, opened 2026-08-11
fix: local file import no longer races the debounced IndexedDB persist

### LINKED ISSUE #665: Local file import fails to display the network on first load (races the debounced IndexedDB persist)

maxkfranz:
Fixes #665.

## Problem

Importing a local network file (Data ▸ Import ▸ Network from File...) adds the network to the in-memory stores and navigates; the URL-driven network swap then reloads the network from IndexedDB. The store persists are debounced (300ms trailing coalescer), so the first load races the write, misses the cache, and `useLoadCyNetwork` classified the just-imported network as a lost local network — showing "Failed to load network data: Local network … is not found in cache" until a manual page reload.

## Fix

`useLoadCyNetwork` now falls back to the in-memory stores on a DB cache miss, before the lost-local-network classification and the NDEx fetch. A new `getCyNetworkFromStores` helper assembles the complete `CyNetwork` (network, tables, view models, active style + style-set snapshot, style options, opaque aspects, undo stack) from the stores.

Two deliberate constraints:

- **Miss-only fallback.** A successful DB read stays authoritative: cross-tab sync (`SyncTabs`) deliberately leaves non-current networks stale in the stores and relies on the network swap re-reading IndexedDB. A memory-first load would resurrect stale copies across tabs. Pinned by a unit test.
- **View models are shallow-copied.** Store state is Immer-frozen and `ViewModelStore.add` mutates the view it is given (viewId/type defaults, selection carry-over), so re-adding the frozen originals would throw. Every other `add` in the reload path was audited for parameter mutation. Pinned by a unit test.

## Tests

- New regression tests in `useLoadCyNetwork.test.ts` (failed with the exact issue error before the fix), plus guards for partial store population and DB-read precedence.
- `test/playwright/local-file-import.spec.ts` now asserts the renderer canvas appears and no "Failed to load network data" message is shown — the gap that let this ship (passes on Chromium against a production build).
- Full unit suite (3456 tests), lint, and typecheck green.

https://claude.ai/code/session_01SF8sXpD2fZmhwPwUbcbDRE

## Summary by CodeRabbit

* **New Features**
  * Added reusable drag-and-drop file upload dialogs with file type and size validation.
  * Added clearer table previews, parsing controls, column editing, and case-sensitive or insensitive joins.
  * Added support for skipping leading lines during table imports.
  * Improved layout processing with additional AntV layout algorithms.
* **Bug Fixes**
  * Improved network loading and cache fallback behavior.
  * Corrected network summary counts and table import edge cases.
  * Preserved multi-line error messages and improved numeric input validation.
* **Performance**
  * Deferred loading of visualization, export, upload, and analysis features for faster startup.


maxkfranz:
@d2fong @chrtannus don't merge this until after #603 as this is branched off of #603

maxkfranz:
@copilot+gpt-5.6-luna:high, merge development into this branch and resolve any conflicts.

PR #677, opened 2026-08-19
Allow a deployment to opt in to installing apps from localhost

keiono:
## What

Lets a deployment opt in to installing apps served from a developer's
`localhost`, and closes the one install path that never checked origins at all.

## Why

An app developer today has to clone and run the host to develop against it. If a
shared host — dev1 — could load an app from their own machine, they would only
need to run their own app. That needs two things from the host: permission to
accept a `localhost` app URL, and an install path that does not already accept
*anything*.

## The opt-in names an origin, not `true`

```json
"allowsLocalhostAppsOn": "https://dev1.ndexbio.org"
```

honoured only when it equals `window.location.origin`.

A boolean would be actively dangerous here. `src/assets/config.json` is the
development server's config — README, "Build for production", says so — and a
production build starts from a copy of it, editing five documented fields. That
list would not have mentioned a new flag, so `true` would reach production
unnoticed. An origin carried forward matches nothing and does nothing.
`installGate.test.ts` asserts that against the real config file, so it fails the
moment anyone converts the field to a boolean.

Default is off: a deployment whose config omits the field behaves exactly as
before.

## The catalog gate turns on provenance, not origin

`activateApp` reached `loadRemoteApp` with no origin check, so a Manifest Source
the *user* set could name any URL and the host would fetch and execute it.

Sending the catalog through `isAllowedOrigin` would have broken the product:
`apps.json` serves every bundled app from `cytoscape.org`, while
`appInstallAllowedOrigins` names `apps.cytoscape.org`. So the rule is
provenance-based — entries from the deployment's own default manifest are the
operator's own list and are trusted; everything else goes through the same gate
as the install paths. An organization's catalog keeps working by being
allow-listed rather than by being unchecked.

## Behaviour change worth a release note

A user who configured a custom Manifest Source pointing at a non-allow-listed
origin will find it stops loading. That is the intended fix, but it is
user-visible and would arrive with a release nobody had flagged as behavioural.

## After this is deployed

The committed config already names `https://dev1.ndexbio.org`, which is dev1's
origin, so **dev1 gains the capability on deploy and nothing else does**. App
developers then run only their own app:

```bash
CYWEB_DEV_HOST=https://dev1.ndexbio.org/cytoscape npm run dev
```

(the app-side half is in `cytoscape-web-app-examples`)

## Testing

- 3541 unit tests. Five separate wirings read the new field; each was
  mutation-checked — the argument removed at that call site, the corresponding
  test confirmed to fail, then restored. Wiring tests pass whether or not they
  assert anything.
- New e2e: a manifest naming `https://blocked.invalid` is refused, asserting
  **no request to that origin was ever made**. Without that assertion "blocked"
  and "failed anyway" are indistinguishable.
- Verified end to end in a real browser against a host on a non-localhost
  origin: with the opt-in matching, the app server received the manifest
  request, then `remoteEntry.js`, then the app's own modules; with it naming a
  different deployment, no confirmation dialog and no `remoteEntry.js` at all.
- **Not** verified: the browser's Local Network Access permission prompt, which
  only appears from a genuinely public origin. Deploying to dev1 is what
  unblocks testing that layer.

## Minor

`config.dev.json` is unreferenced by any code and has drifted from
`config.json`. Removing it was the plan, but it carries a
`debugOptions.disableAutoReload` flag that exists nowhere else and that nothing
reads. That flag is now also in `config.json` so it survives either decision;
the file is left in place pending your call on it.

## Summary by CodeRabbit

* **New Features**
  * Added deployment-specific opt-in support for installing apps served from localhost.
  * Localhost app installation, custom manifest URLs, workspace restoration, and app activation now respect configured origin rules.
* **Bug Fixes**
  * Blocked apps from unapproved origins are no longer loaded and are clearly reported.
  * Improved handling of failed or disallowed app activations.
* **Documentation**
  * Documented localhost installation configuration and deployment requirements.
* **Tests**
  * Added coverage for allowed, blocked, and mismatched deployment scenarios.


keiono:
Thanks — three of the five findings held up under checking and are fixed in baec63f8.

**Missing catalog provenance was trusted** (Copilot, CodeRabbit). `catalogSources[id] ?? 'manifest'` labelled an entry with no recorded provenance as coming from the deployment's own manifest — exactly the case the gate exists to distinguish. The store writes a provenance for every entry today, but the catalog is persisted and restored, and a partial restore must not read as "the operator's own list". The parameter is now optional and anything not explicitly `'manifest'` goes through the origin check.

**IPv6 loopback was refused** (Codex). Confirmed: `new URL('http://[::1]:6000/x').hostname` is `"[::1]"`, brackets included, so neither the host-side nor the URL-side check matched an address that is loopback by definition. The three copies of the two-hostname comparison are now one predicate that includes it.

**`isLocalhostAppOptIn` had a dishonest signature** (Copilot). It declared `string | undefined` while deliberately handling non-string values from unvalidated JSON, making that branch unreachable by type and forcing a cast in the test. It takes `unknown` now, which is what it is.

Each fix was mutation-checked rather than assumed: reinstating the provenance fallback fails one test, removing IPv6 from the loopback set fails three.

**Not taken:** converting three src-rooted imports to the `@/` alias (CodeRabbit). Every sibling import in those files is relative, one of the three lines is not part of this PR, and a single aliased import among four relative ones reads worse than either convention applied consistently. Happy to do it as a separate sweep if you would rather the whole files moved.

keiono:
Good catch, and it is worse than the report suggests — fixed in b40a1f99.

`reconcileInstalledStatus` stamps `source: 'manifest'` on an app's first successful activation **whichever manifest it came from**, and `composeCatalog` hands an installed-only entry its stored source straight back. So an app activated from a custom Manifest Source persists as `'manifest'`; reset the Manifest Source to default and the old test passed, loading an unlisted app with no origin check.

The same shape also defeated the workspace-restore gate added earlier in this PR. A snapshot carrying an unlisted app is imported inactive with its source preserved — and activating it afterwards took the exempt path, so importing it inactive bought nothing.

The exemption now asks the question it always meant to ask: **does the deployment's own default manifest list this entry, right now?** Answered against the manifest as loaded, and matched on **url as well as id** — `composeCatalog` lets an installed `appstore`/`snapshot` entry win a collision with a manifest entry of the same id, so id alone would vouch for a URL the manifest never named.

`AppSource` no longer participates, which retires the "missing provenance is trusted" fix from the previous commit: there is no provenance left to be missing.

Mutation-checked — putting the provenance test back reproduces the hole and fails two tests. Also verified by hand, since no automated test covers it and this change could have broken every bundled app: with the host serving its default catalog, activating an app from it still loads `remoteEntry.js` and evaluates the app module, with no refusal.

PR #689, opened 2026-08-27
feat(benchmark): port the cytoscape.js v4 benchmark harness (#601)

### LINKED ISSUE #601: Performance: benchmarks for loading and editing networks
## Purpose

We have no benchmarks and effectively no timing instrumentation anywhere in the app, so we can't answer "did this PR make network loading slower?" or "how long does editing a 20k-element network actually take?" except by feel. Some repeatable way to measure loading and editing would give the performance work something to stand on.

## Context

- `logPerformance` is defined in `src/debug.ts:88` but has **zero call sites**. The only runtime timing in the entire codebase is a 100ms guard on plugin `mount()` in `src/data/hooks/stores/appLifecycle.ts:41-49`. No `performance.mark/measure`, no timing around network load, CX2 parsing, DB writes, or layouts.
- No `bench`/`benchmark`/`perf` scripts exist anywhere in `package.json`, `scripts/`, or `test/`.
- Vitest 4 is already in place, so `vitest bench` is available with no new dependencies — though it runs in jsdom + fake-indexeddb, which won't reflect real rendering or real IndexedDB latency. A real-browser pass (Playwright timing, DevTools tracing) measures different things; they're complements, not substitutes.
- Fixture infrastructure exists but isn't perf-oriented: `scripts/generate-test-fixtures/` (CX2/SIF/table generators) and `scripts/download-ndex-networks.ts` (pull real networks from NDEx). Current `test/fixtures/` files are small correctness fixtures. Real NDEx networks have attribute/style density that synthetic ones miss.
- Size thresholds already baked into the app, useful for bracketing fixture sizes: `maxNetworkElementsThreshold: 26000` and `maxEdgeCountThreshold: 20000` (`src/assets/config.json`), and the layout-selection cutoff `ELE_THRESHOLD = 1000` (`src/models/LayoutModel/impl/layoutSelection.ts:39`).
- One perf-relevant behavior we already know about: every store mutation re-serializes (main-thread `structuredClone` via `toPlainObject`) and rewrites the full network slice to IndexedDB — `putTablesToDb` even serializes twice. Bulk edits and node dragging cause write storms. Editing measurements would make this cost visible. (Details in `scratch/ticket_themes/Data_Persistence_and_Architecture_Robustness.md`, "No write batching/debouncing".)

## Questions

- Which harness? `vitest bench`, Playwright timing, a custom script, some mix of them
- Fixture strategy: generated synthetic sizes, pinned real NDEx networks, both? Commit them or fetch on demand (they can be large)?
- Where should numbers live to be useful — a committed baseline doc, CI artifact, nowhere-formal-yet?
- Is CI variance even acceptable for timing, or are benchmarks a local-only tool for now?

## Possible Directions

Operations that seem worth measuring, in rough order of how often users feel them — pick whatever slice interests you:

- **Loading**: CX2 JSON → model conversion (`CxModel` converters, `validateCX2`); the full parse → populate stores → write-to-IndexedDB path; and the cold read-back path (`getCyNetworkFromDb`).
- **Editing**: add/delete nodes and edges (including the persist write they trigger); bulk attribute edits through `TableStore` (the re-serialize hotspot); undo/redo on large networks (`useUndoStack`/`UndoStore`); visual style updates that regenerate views.
- `CyJsRenderer` is where the computed viewmodels are transformed into cyjs json objects

- A benchmark suite of any size — even one repeatable measurement of one operation is more than we have.
- Graded fixtures others can reuse (perf work, e2e tests, demos).
- A baseline document of first numbers.
- Or just a harness spike and your recommendation on what's worth building.

## References

- `src/debug.ts:88` — the unused `logPerformance` namespace
- `src/models/CxModel/` — parse/convert; `src/data/db/index.ts` — `putNetworkToDb`, `getCyNetworkFromDb`
- `scripts/generate-test-fixtures/`, `scripts/download-ndex-networks.ts`
- https://github.com/cytoscape/cytoscape-web/tree/development/src/features/NetworkPanel/CyjsRenderer -- this folder is the main module where cytoscape webs models are transformed into cytoscape.js json.  There is a react loop runs these functions every time the vizmapper/table data/network data has changed

maxkfranz:
**WIP / preview — opened for early review.** Closes #601.

The cytoscape.js v4 benchmark harness is ported to cytoscape-web and verified end to end, across five commits. Nothing in `src/` changes behavior; this is harness, suites, tests, and docs, plus two latent script bugs the port surfaced.

## What landed

**Harness core (`benchmark/`)** — mitata-timed suites, run one-process-per-job by `report.mjs`, with the v4 machinery intact: the `BENCH_N` / `BENCH_OP` / `BENCH_JSON` env protocol, `--repeat 3` per-row median merging with `repeatSpread` noise bands, `--jobs` concurrency (which gets its own epoch), machine fingerprints, and harness fingerprints so a comparison refuses to cross a methodology change. New devDeps: `mitata` + `tsx`, as approved.

**Six Node suites** covering the loading and editing asks in the issue:

| suite | covers |
| --- | --- |
| `load` | validate / convert / per-aspect / export, plus the committed 2.5 MB NDEx network as a real-density row |
| `serialization` | the `toPlainObject` + Map-serializer path |
| `table-edit` | table mutation cost |
| `network-edit` | including the live `nodes`/`edges` getter cost `putNetworkToDb` silently pays |
| `render-transform` | the `CyjsRenderer` pipeline with real passthrough / discrete / continuous mappings |
| `db` | under `fake-indexeddb` — `putTablesToDb`'s triple-serialize has a named tripwire row |

**Browser pass** (`npm run benchmark:browser`, or `--renderer`) — boots the built `dist/` headless in Chromium from an ephemeral-port server, deliberately never touching the e2e stack's port 5500, reads the app's own `cyweb.boot.*` milestones, and samples a real `CyWebApi` CX2 import. 2000 nodes imports in ~53 ms warm; boot reaches `workspace-hydrated` at ~320 ms.

**Publish / compare** — seeded `benchmark/published/` archive, with the dirty-tree and concurrent-run refusals verified live. `benchmark:compare` renders per-machine comparison pages; the two-run smoke test showed 19 shared rows, −0.5 % drift, zero false movers.

**Tests and docs** — 34 node-env unit tests over the pure modules (median high-side merging, the epoch ledger, the mixed-fingerprint throw, control-twin lookup, …), `docs/agents/benchmarking.md` with the methodology rules, AGENTS.md pointers, and agent allowlist entries.

## Adaptations worth knowing

- The report renders single labelled rows instead of v3-vs-gpu dumbbells, and the scaling table shows p50-vs-N rather than speedups.
- There is no v3 to pair against, so each group's **frozen control row** (structuredClone-class work) serves as the per-row noise twin.
- Two porting traps are recorded in `lessons.md`: `tsx` can't named-import through `export *` barrels, and a `styleEnabled` cytoscape instance keeps the process alive without `process.exit(0)`.
- Fixed two latent bugs `tsc` and the port surfaced: `generate-cx2.ts`'s unnarrowed optional param, and `download-ndex-networks.ts`'s stale import.

## Gates

- `npm run test:checks:quiet` green — 3828 tests
- `npm run benchmark -- --repeat 2` completes in 2.8 min, 62/62 rows carrying noise bands, zero failures

## Not done yet

No baseline is published, per your call. When you want one:

```bash
npm run benchmark:all -- --repeat 3   # on a clean tree
npm run benchmark:publish -- --note "first baseline"
```

https://claude.ai/code/session_014kAQhEuoFo3PznQyy1S847


maxkfranz:
@d2fong @chrtannus, are there other areas that should be benchmarked?  There's only 6 basic ones for now.

Issue #628, opened 2026-07-23
Dialogs: backdrop-click and Esc dismissal is inconsistent across 43 modal surfaces, with no documented policy


coleslaw481:
## Summary

Dialog dismissal behavior (backdrop/clickaway click and <kbd>Esc</kbd>) is decided per call site across ~43 modal surfaces, with no documented convention and no shared wrapper. **35 of 43 live modal surfaces dismiss on both backdrop click and <kbd>Esc</kbd>; 8 dismiss on neither.** The Apps → Manage Apps dialog named in the original report is one of the 8.

The deliverable here is not a one-line fix: it is (a) a written policy, (b) a mechanism that can express it, and (c) a sweep. A full audit table is included below so the sweep does not have to be re-derived.

## Reported behavior

Apps → Manage Apps can only be closed via its **Close** button — clicking outside does nothing — which is inconsistent with other dialogs in the app. Reported on `dev1.ndexbio.org/cytoscape`, Version 1.0.7, Build `d9fd128-07-22-2026-11-35-44`. Screen recording in the original report at the bottom.

**The report is correct.** `AppSettingsDialog.tsx:234` passes no `onClose` prop, so neither backdrop click nor <kbd>Esc</kbd> can dismiss it. Still reproduces on `development`.

Two claims from the discussion thread need correcting:

- **"It's actually a minority of dialogs that can be closed by clickaway"** ([comment](https://github.com/cytoscape/cytoscape-web/issues/628#issuecomment-5061479585)) — this is backwards. Dismissable is the **majority** (35 of 43). The recollected decision to "make all dialogs not closeable by clickaway" was never written down and never implemented at scale; only PR #249 applied it, and only to the two Vizmapper/Network-property popovers.
- **"It's not clear why"** some dialogs close on <kbd>Esc</kbd> and others don't ([comment](https://github.com/cytoscape/cytoscape-web/issues/628#issuecomment-5107415586)) — now answered, see Root cause. It is not intentional; it is a side effect of whether an author happened to pass `onClose`.

## Root cause

Three findings, all confirmed by reading the code and the installed MUI 5.18.0 source:

[...]

PR #695, opened 2026-08-31
Add the 'modal-launcher' resource slot so apps can open host-themed dialogs

### LINKED ISSUE #690: Add a `modal-launcher` slot so apps can open host-themed dialogs without owning a React root
## Problem: apps without a persistent surface can't open host-integrated UI

CW app's components render inside the host's React tree only while a host-managed slot renders them: a `right-panel` tab, an `apps-menu` item while the dropdown is open, or a `search-bar` provider's `optionsComponent` while the **More Options** popover is open (see #688). An app whose only registration is a search provider has **no mounted component at all** most of the time — yet its `onSubmit` callback may need to open UI at exactly such a moment.
The concrete case: [PSICQUIC's port](https://github.com/cytoscape/psicquic-cw) of the desktop "Select Databases" dialog, which must open when the count phase finishes, gating the import on the user's source selection:

```TypeScript
export const PsicquicApp: CyAppWithLifecycle = {
  id,
  name: displayName,
  description,
  version,
  apiVersion: '1.0',

  resources: [
    {
      slot: 'search-bar',
      id: 'psicquic-search',
      name: 'PSICQUIC',
      description: 'Search PSICQUIC-compliant databases.',
      website: 'http://psicquic.github.io/',
      icon: psicquicLogo,
      placeholder: 'Enter gene/protein IDs (e.g. brca2)...',
      optionsComponent: lazy(() => import('./components/SearchOptionsPanel')),
      // onSubmit reaches the App API through the module-level context set in mount().
      // runPsicquicSearch will open a Dialog component.
      onSubmit: ({ query }) => runPsicquicSearch(query),
    },
  ],

  mount(context: AppContext): void {
    setAppContext(context)
    // The Select Databases modal needs a render surface that outlives the
    // options popover — a small app-owned React root on document.body.
    mountDialogHost()
  },

  unmount(): void {
    // Resources are auto-cleaned by the host on deactivation; the dialog
    // root and the parked context are ours to drop.
    unmountDialogHost()
    clearAppContext()
  },
}
```

The only way to do this today is `ReactDOM.render` into an app-owned `<div>` on `document.body` — a second, independent React tree. That works (`psicquic-cw` ships it), but every consequence below follows from it:

1. **The host's MUI theme doesn't reach the dialog.** React context doesn't cross roots, so the host's `CssVarsProvider` never reaches the app's tree and the dialog renders in MUI's default light theme — visibly broken in dark mode. The workaround is a hand-maintained copy of the host theme, which silently drifts whenever the host theme changes, plus two non-obvious traps: the copy must use a unique `cssVarPrefix` (both roots share the host's Emotion cache, so a second provider emitting the default `--mui-*` variables overrides the *host's* palette document-wide), and styling must use theme tokens rather than `theme.palette.*` (which holds light-scheme literals under a CSS-variables theme).
2. **The app is forced onto the deprecated `ReactDOM.render`.** The Module Federation share map covers root specifiers only, so importing `react-dom/client` for `createRoot` bundles a second react-dom and fails the SDK's no-shared-payload gate. Legacy render works on React 18 but is removed in React 19 — when the host bumps its React major, every app with a satellite root breaks.
3. **Host UI policies can't be enforced.** The host guarantees its dialog conventions (issue #628's dismissal policy) structurally via `CyDialog` plus lint — neither reaches third-party bundles. Each app reinvents dismissal, and a policy change (e.g. the planned Escape-to-close standardization) must be re-implemented app by app instead of once in the host.
4. **No error isolation.** Slot components get `PluginErrorBoundary` + `Suspense` + the host's `PluginFallback`; a crash in an app-owned root is invisible to the host and takes the app's UI down silently.
5. **No lifecycle management.** The host auto-cleans slot resources on app deactivation; a foreign root must be manually mounted/unmounted by the app (`mount()`/`unmount()`), and a forgotten cleanup leaks DOM and React trees.
6. **Two uncoordinated modal systems.** The host can't know an app modal is open — z-index/stacking against host dialogs, onboarding overlays, focus traps, and keyboard shortcuts are all unmanaged.

The same root cause bites beyond search providers: an `apps-menu` item that wants a confirmation or parameters dialog renders it from a component that unmounts when the dropdown closes ([network-analyzer-cw](https://github.com/cytoscape/network-analyzer-cw)'s menu item already carries a workaround for exactly this unmount-kills-work behavior).

## Solutions considered

- **A. Status quo** — app-owned root + theme copy. Ships today; suffers all six problems, duplicated per app.
- **B. Expose the host theme through the app API** (e.g. a `cyweb/Theme` expose or `apis.ui.getTheme()`). Fixes only problem 1; the satellite root, legacy render, policy, isolation, and lifecycle issues all remain.
- **C. `companionComponent` on `RegisterNetworkSearchProviderOptions`** — an always-rendered hidden component per search provider, inside the host tree. Additive and cheap; fixes 1–5 — but only for the `search-bar` slot, and every future "headless" surface re-raises the same question.
- **D. The `modal-launcher` slot** — already reserved in the `ResourceSlot` docs. General: any app registers modal content; the host renders the shell inside its own tree.
    ```TypeScript
    /**
     * Identifies a specific host-managed UI location that plugins can occupy.
     *
     * Current slots:
     *   'right-panel'  — tabbed side panel on the right
     *   'apps-menu'    — dropdown in the Apps toolbar button
     *   'search-bar'   — network search bar at the top of the Workspace tab
     *
     * Reserved for future rollouts:
     *   'left-panel', 'bottom-panel', 'tools-menu', 'status-bar', 'modal-launcher'
     */
    export type ResourceSlot = 'right-panel' | 'apps-menu' | 'search-bar'
    ```

## Why `modal-launcher`

- **Theming is solved structurally, not by copying.** The content renders under the host's `CssVarsProvider`; MUI's `Dialog` portal moves only DOM, and context follows the component tree (this is exactly why MCODE's `NewAnalysisDialog`, rendered from a `right-panel` component, themes correctly with zero effort).
- **No second React root** — the legacy `ReactDOM.render` hack and the share-map constraint disappear, and apps survive the host's next React major.
- **One place for modal policy.** The host wraps content in its `CyDialog` shell, so the dismissal rules — including the Escape-to-close change under discussion — apply to every app modal by changing the host once.
- **Same guarantees as every other slot**: `AppIdProvider` (so `useAppContext()` works), `PluginErrorBoundary` with `errorFallback`, `Suspense` for lazy content, upsert semantics, introspection via `getRegisteredResources`, and auto-cleanup on app deactivation.
- **Host-coordinated modality**: stacking, focus, and "a modal is open" awareness (tours, shortcuts) in one modal system.
- **It generalizes**: search providers, menu actions, context-menu handlers, and future background/headless apps all get a sanctioned way to open dialogs.

### Sketch of the contract

```ts
// Registration (declarative in `resources` or via apis.resource):
{ slot: 'modal-launcher', id: 'select-databases',
  maxWidth?: 'sm' | 'md' | ..., component: ComponentType<ModalHostProps> }

// Injected props — the app's component renders the DialogTitle/Content/Actions
// children inside the host's CyDialog shell:
interface ModalHostProps { requestClose: () => void }

// Imperative open, since modals are launched from app logic (e.g. onSubmit):
apis.resource.openModal('select-databases')
```

All the implementation ingredients already exist in the host: the `resourceApi` validator, `RegisteredAppResource` storage, the slot wrapper stack, and `CyDialog`.

### Immediate payoff in existing apps

[psicquic-cw](https://github.com/cytoscape/psicquic-cw) deletes `dialogHost.tsx` and the entire `theme.ts` copy; `SelectDatabasesDialog` becomes a registered modal opened from `onSubmit`. network-analyzer-cw's menu item could open its analysis form without the dropdown-lifetime workaround.

chrtannus:
Closes #690.

## What

Apps can now register modal dialog content that the host renders **inside its own React tree**, and open it imperatively from app logic that has no mounted component — a search provider's `onSubmit`, an apps-menu action:

```ts
// Declarative (in `resources`) or via apis.resource.registerModal(...):
{ slot: 'modal-launcher', id: 'select-databases', maxWidth: 'md', fullWidth: true,
  component: lazy(() => import('./SelectDatabasesDialog')) }

// Imperative open/close, bound to the app's identity:
apis.resource.openModal('select-databases')   // idempotent; RESOURCE_NOT_FOUND if unregistered
apis.resource.closeModal('select-databases')
```

The app's component renders the DialogTitle/Content/Actions children and receives `ModalHostProps { requestClose }`; the host owns the `CyDialog` shell.

## How

- **Types/model**: `ResourceSlot` gains `'modal-launcher'` (app-api union + the model-layer duplicate), with `RegisterModalOptions`, `ModalHostProps`, and `registerModal` / `unregisterModal` / `openModal` / `closeModal` on `ResourceApi`. Types flow to `cyweb/ApiTypes` and `@cytoscape-web/api-types` automatically (no new federation expose).
- **Open state**: new `ModalLauncherStore` (a stack of `{appId, id}`, idempotent open/close), written from the React-free core API. It self-registers with `AppCleanupRegistry`, so open modals auto-close on app deactivation or mount failure; `unregisterModal` and `unregisterAll` also close what they remove.
- **Renderer**: `ModalLauncherHost`, always mounted in `AppShell`, renders one `CyDialog` per open entry with the standard slot wrapper stack — `AppIdProvider` → `PluginErrorBoundary` → `Suspense` (spinner fallback for large lazy chunks) — plus a **structural host-rendered Close "X"** sharing the `requestClose` path. That enforces the dialog-dismissal policy (#628) for every app modal at once: backdrop/Esc inert, and a guaranteed exit even when the app content crashes into its error fallback.
- Because open state lives in a store and the dialog renders under `AppShell`, a modal **survives the unmount of whatever launched it** (a closing apps-menu dropdown, the search options popover) and route changes.

## Why the host X (design note)

`dialogPolicy.test.ts` statically guarantees every `CyDialog` an exit, but it cannot see third-party bundles — the host-rendered X makes the policy structural for app modals instead of relying on each app. Apps still wire their own Cancel/Done buttons to `requestClose`.

App authors must **not** treat component unmount as a "modal closed" signal: the host runs under `React.StrictMode`, whose dev double-mount runs effect cleanups once right after mount. Documented in `Api.md` (found the hard way in psicquic-cw's migration).

## Testing

- Unit: registerModal/openModal/closeModal/unregisterModal coverage in `resourceApi.test.ts` (incl. appId isolation), `ModalLauncherStore.spec.ts` (incl. the `cleanupAllForApp` path), `ModalLauncherHost.spec.tsx` (render, both close paths, stale-entry/inactive-app filtering, width forwarding, error isolation, `React.lazy`), and first-ever coverage for `processDeclarativeResources`. `npm run test:checks:quiet`: 315 files / 3895 tests green.
- E2E: the fixture remote registers a modal plus a `closeOnAction: true` menu item that opens it — `remote-app-load.spec.ts` proves the modal outlives the dropdown that launched it; a new `dialog-dismiss.spec.ts` row proves Esc/backdrop are inert and the host X exits.
- Manually verified against both real consumers (psicquic-cw's Select Databases dialog through a full search→count→import run, and network-analyzer-cw's analyzer form from both its launchers); their migrations land separately in their own repos.

## Also in this PR

- Docs: `Api.md` ResourceApi section, `DIALOG_DISMISS_POLICY.md`, design docs' reserved-slot lists.
- `@cytoscape-web/api-types` stays at `1.0.0-beta.4` (never published); the unpublished beta.4 changelog entry absorbs the modal-launcher types and the previously-unlogged search-bar types.

## Summary by CodeRabbit

* **New Features**
  * Added support for host-rendered modal dialogs from apps.
  * Modals can be opened, closed, configured for size, and remain open after the launching UI closes.
  * Added loading states, error handling, and a consistent host-provided Close button.
  * Modals automatically close when their app is deactivated or removed.

* **Documentation**
  * Updated API documentation, specifications, changelog, and roadmap to describe modal support and dismissal behavior.

* **Tests**
  * Added unit and end-to-end coverage for modal registration, lifecycle, rendering, dismissal, and app isolation.


chrtannus:
> In `@src/features/AppManager/ModalLauncherHost.tsx` at line 53, Update the React
key in the modal launcher rendering to use JSON.stringify([appId, id]) instead
of delimiter concatenation, ensuring distinct app and modal identifiers always
produce unambiguous keys.

**Skipped — JSON.stringify React key** (`ModalLauncherHost.tsx`): the key `${appId}::modal-launcher::${id}` is the repo-wide identity convention — every `resourceId` is `appId::slot::id`, and the AppMenu renderer keys its items with the identical format (`AppMenu/index.tsx:101,118`). A collision would require :: inside an appId, which is a Module Federation container name from the app registry, not arbitrary input. Switching this one site to `JSON.stringify` would break convention without addressing the pattern everywhere else it exists.

PR #701, opened 2026-09-01
fix(app-manager): stop offering Uninstall for apps.json apps (#699)

### LINKED ISSUE #699: App Manager: apps.json apps stay uninstallable after a snapshot restore or App Store install
<details>
<summary>Original request (2026-08-31)</summary>

If an app is loaded from the app.json, we should disable the deletion of it in the CyWeb UI. Update and disable should still be allowed.

</details>

---

## Summary

The App Manager hides **Uninstall** for `apps.json` apps only when the app has never been installed another way. Two paths — workspace snapshot restore and App Store / Install-from-URL — stamp a non-manifest provenance on an id that also ships in `apps.json`, which puts the kebab menu and **Uninstall** back on the row.

## Expected behavior

An app whose id is present in the resolved manifest (`apps.json`, or a custom manifest URL) shows the enable/disable toggle only. No kebab menu, no **Uninstall**. Enable, disable, and catalog refresh stay available.

This is what [`workspace-app-install-design.md` §12.3](docs/design/module-federation/specifications/workspace-app-install-design.md) already specifies: *"`manifest` | No — disable only | Removal is meaningless; `Refresh` would re-add it from `apps.json`"*.

## Actual behavior

The row shows a kebab menu with **Uninstall** whenever `catalogSources[id]` is `appstore` or `snapshot`, regardless of manifest membership. Confirming the dialog removes the workspace record, then `recomposeCatalog()` re-adds the manifest entry, so the row reappears as a disable-only manifest app. The dialog's text — *"It will be removed from this workspace"* ([AppListPanel.tsx:331](src/features/AppManager/AppListPanel.tsx#L331)) — does not hold.

Two ways to reach that state:

| Path | Where the source is stamped |
| --- | --- |
| Restore a saved workspace that had an `apps.json` app installed | [useLoadWorkspace.ts:154](src/data/hooks/useLoadWorkspace.ts#L154) — `source: 'snapshot'` for every restored entry |
| Install an app from the App Store or `?installApp=` whose id is in `apps.json` | [useAppManager.ts:372](src/data/hooks/stores/useAppManager.ts#L372) — `source: 'appstore'` for every install |

## Root cause

Removability keys off the installed record's provenance, not off manifest membership. `removable: source === 'appstore' || source === 'snapshot'` ([AppListPanel.tsx:113](src/features/AppManager/AppListPanel.tsx#L113)).

On an id collision, `composeCatalog` lets the installed record win and overwrites the `'manifest'` tag with `'appstore'`/`'snapshot'` ([composeCatalog.ts:46](src/features/AppManager/manifest/composeCatalog.ts#L46)). Both stamping sites above apply their source unconditionally, so manifest membership is lost from `catalogSources` and the panel has no other signal to consult.

Uninstall is then a partial no-op: `uninstallApp` drops the workspace record ([useAppManager.ts:393](src/data/hooks/stores/useAppManager.ts#L393)) and `recomposeCatalog` re-adds the manifest entry from `manifestEntriesRef` ([useAppManager.ts:140](src/data/hooks/stores/useAppManager.ts#L140)). Confirmed by reading the code; not reproduced in a running build.

## Scope of change

| File | Change |
| --- | --- |
| `src/features/AppManager/manifest/composeCatalog.ts` | Return manifest membership per id alongside `entries` and `sources`. It is the only place that sees both inputs. |
| `src/models/StoreModel/AppStoreModel.ts` | Add the membership map to `AppState`; extend the `setCatalog` signature. |
| `src/models/StoreModel/impl/appStoreImpl.ts`, `src/data/hooks/stores/AppStore.ts` | Store the map in `setCatalog`. |
| `src/features/AppManager/AppListPanel.tsx` | Compute `removable` as `!inManifest && (source === 'appstore' \|\| source === 'snapshot')`. Decide the chip label for a pinned install over a manifest id (line 215). |
| `src/data/hooks/stores/useAppManager.ts` | Guard `uninstallApp` against manifest ids, so the command is safe when called outside the panel. |
| `docs/design/module-federation/specifications/workspace-app-install-design.md` §12.3 | The table keys removability on `source`; restate it as manifest membership. |
| `src/features/AppManager/manifest/composeCatalog.test.ts` | Extend for the collision case. |
| `src/features/AppManager/AppListPanel.spec.tsx` | New file — the panel has no test today. |

## Proposed approach

Gate removability on manifest membership rather than on `InstalledApp.source`. `composeCatalog` already computes `inManifest` internally at [composeCatalog.ts:35](src/features/AppManager/manifest/composeCatalog.ts#L35) — return it, thread it into `AppStore`, and read it in `AppListPanel`. `catalogSources` keeps its current meaning (which URL won) and is no longer overloaded as the removability signal.

The one contested case is an App Store install pinned over a manifest id:

| Option | Cost |
| --- | --- |
| Hide Uninstall entirely (recommended) | The user cannot revert a pinned version back to the manifest URL from the UI. |
| Keep the menu, relabel to **Revert to manifest version** | Matches what the code actually does, but adds a second action and a second confirmation string. |

## Open questions

- Should a pinned App Store version over a manifest id keep a **Revert to manifest version** action, or nothing at all?
- Should `uninstallApp` itself refuse manifest ids, or is hiding the UI affordance sufficient?

## Verification

1. `src/features/AppManager/manifest/composeCatalog.test.ts` — an id in both the manifest and `installedApps` with `source: 'snapshot'` reports `inManifest === true`.
2. New `src/features/AppManager/AppListPanel.spec.tsx` — render with `catalog` holding a manifest id and `catalogSources[id] === 'snapshot'`; assert `app-kebab-<id>` is absent and `app-toggle-<id>` is present. Assert the inverse for an id absent from the manifest.
3. `src/data/hooks/stores/useAppManager.test.tsx` — `uninstallApp('hello')` on a manifest id leaves `workspace.installedApps` unchanged.
4. Manual: activate `hello` from `apps.json`, save the workspace, restore it in a fresh profile, open App Manager, confirm the row has the toggle and no kebab.

d2fong:
Fixes #699.

## What was wrong

Removability keyed off `InstalledApp.source` (`AppListPanel.tsx:113`), but `composeCatalog` lets a pinned `appstore`/`snapshot` record win an id collision and overwrite the `'manifest'` tag (`composeCatalog.ts:46`). Both stamping sites apply their source unconditionally:

| Path | Site |
| --- | --- |
| Restore a workspace snapshot | `useLoadWorkspace.ts:154` — `source: 'snapshot'` for every restored entry |
| App Store / `?installApp=` | `useAppManager.ts:372` — `source: 'appstore'` for every install |

So an app shipped in `apps.json` got the kebab menu and **Uninstall** back. Confirming it discarded the pinned URL, then `recomposeCatalog` re-added the manifest entry — the row returned and the dialog's "It will be removed from this workspace" did not hold.

## The fix

Gate removability on manifest membership, not on `source`.

- `composeCatalog` returns `manifestIds` alongside `entries` and `sources`.
- `AppStore.manifestIds` holds it; `setCatalog(entries, sources?, manifestIds?)` falls back to the `'manifest'`-sourced ids when the argument is omitted, so existing two-argument callers keep their behavior.
- `AppListPanel` hides the kebab for any id the manifest still ships.
- `uninstallApp` refuses those ids, guarding every caller outside the panel.

The source chip now tracks `source` rather than removability. A pinned install shadowing a manifest id keeps its `App Store` / `Snapshot` chip — otherwise losing the kebab would also lose the only signal that the row runs a pinned URL.

## Cost

A pinned App Store version installed over a manifest id can no longer be reverted to the manifest version from the UI. That was never a labelled action — it was an **Uninstall** that appeared to fail — so nothing discoverable is lost. Spec §12.3 now records `Revert to manifest version` as the place to add it back if wanted.

This resolves both open questions on the issue: hide the affordance entirely, and guard the command as well as the UI.

## Tests

| File | Coverage |
| --- | --- |
| `AppListPanel.spec.tsx` (new) | 9 cases: kebab hidden for manifest / snapshot-shadowed / appstore-shadowed rows, shown for non-manifest rows, full uninstall flow, chip behavior |
| `composeCatalog.test.ts` | `manifestIds` under both collision sources |
| `appStoreImpl.test.ts` | `setCatalog` fallback and explicit `manifestIds` |
| `useAppManager.test.tsx` | `uninstallApp` refuses a manifest id |

The new panel spec fails 3/9 against the old `removable` rule and passes against the fix.

`npm run test:checks:quiet` — lint, tsc, 317 files / 3922 tests, all passing. No e2e spec touches the uninstall affordance; port 5500 had a dev server running, so no local e2e run.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented uninstalling apps provided by the workspace manifest.
  * Preserved manifest-provided apps when App Store or snapshot installations use the same ID.
  * Updated app list controls to show uninstall options only when removal is supported.

* **UI Improvements**
  * Source indicators now accurately show whether an app comes from the App Store or a snapshot, including apps overriding manifest entries.

### DIFF
--- a/docs/design/module-federation/specifications/workspace-app-install-design.md
+++ b/docs/design/module-federation/specifications/workspace-app-install-design.md
@@ -780,11 +780,11 @@ orphan `removeOrphan` remains unconfirmed; only true uninstall is confirmed.)
 
 | File | Change |
 | --- | --- |
-| `AppListPanel.tsx` | `AppDisplayEntry` gains `source` and `removable`; `getAction()` is split into a primary-action selector and a `removable` predicate; add the overflow menu and the confirmation dialog |
+| `AppListPanel.tsx` | `AppDisplayEntry` gains `source` and `removable`; `getAction()` is split into a primary-action selector and a `removable` predicate; add the overflow menu and the confirmation dialog. `removable` reads `AppStore.manifestIds`, not `source` (§12.3) |
 | `AppSettingsDialog.tsx` | Add the **Install from URL** action on the Apps tab (§12.8), distinct from the existing manifest-source controls |
 | `useAppManager.ts` | Implement `uninstallApp(id)`; add `installApp`/`uninstallApp` to `AppManagerCommands` |
 | `WorkspaceStore.ts` | `removeInstalledApp(id)` / `setInstalledAppStatus(id, status)` actions (§10.2) |
-| catalog merge (§8.1) | Propagate `source` onto each merged catalog entry so the panel can decide removability |
+| catalog merge (§8.1) | Return `manifestIds` alongside `entries` and `sources`. The panel decides removability from `manifestIds` (§12.3); `source` only drives the provenance chip |
 
 `AppManagerCommandsContext` needs no change — it spreads the hook's command
 surface, so the new commands flow through automatically.
--- a/docs/design/module-federation/specifications/workspace-app-install-design.md
+++ b/docs/design/module-federation/specifications/workspace-app-install-design.md
@@ -709,15 +709,34 @@ Two distinct operations must be visually separated:
 Uninstall removes the app from the workspace and therefore from any future
 workspace snapshot, so it is treated as destructive and requires confirmation.
 
-### 12.3 Removability by source
+### 12.3 Removability by manifest membership
 
-Whether a row can be uninstalled is driven by `InstalledApp.source`:
+Whether a row can be uninstalled is driven by **manifest membership**, not by
+`InstalledApp.source`:
 
-| Source | Uninstall? | Rationale |
+| Row | Uninstall? | Rationale |
 | --- | --- | --- |
-| `manifest` | No — disable only | Removal is meaningless; `Refresh` would re-add it from `apps.json` |
-| `appstore` / `snapshot` | Yes | Lives in `workspace.installedApps`; removing it is a real uninstall |
+| id present in the resolved manifest | No — disable only | Removal is meaningless; `recomposeCatalog` re-adds it from `apps.json` |
+| id absent from the manifest, `source: appstore` / `snapshot` | Yes | Lives only in `workspace.installedApps`; removing it is a real uninstall |
 | orphan (legacy) | Yes (fallback) | Existing `removeOrphan` retained for pre-migration leftovers |
+
+`source` alone is the wrong test. On an id collision `composeCatalog` (§8.1)
+lets a pinned `appstore`/`snapshot` record win and overwrite the `'manifest'`
+tag, and both stamping sites apply their source unconditionally: a snapshot
+restore tags every restored entry `'snapshot'`, an App Store install tags every
+install `'appstore'`. Keying off `source` therefore offered **Uninstall** on
+`apps.json` apps, and confirming it only discarded the pinned URL before the
+manifest entry returned (issue #699).
+
+`composeCatalog` returns `manifestIds` alongside `entries` and `sources`;
+`AppStore.manifestIds` holds it, `AppListPanel` reads it, and `uninstallApp`
+refuses any id it contains.
+
+**Cost:** a pinned App Store version installed over a manifest id can no longer
+be reverted to the manifest version from the UI. Reverting was never labelled
+as such — it was an **Uninstall** that appeared to fail — so nothing
+discoverable is lost. A dedicated `Revert to manifest version` action is the
+place to add it back.
 
 ### 12.4 Row layout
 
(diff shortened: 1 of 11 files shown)


d2fong:
@coderabbitai Thanks — the §12.6 finding was correct and is fixed in c4bda55.

**Stale removability rows in §12.6 — fixed.** Two rows contradicted §12.3, not one:

- `catalog merge (§8.1)` now reads "Return `manifestIds` alongside `entries` and `sources`. The panel decides removability from `manifestIds` (§12.3); `source` only drives the provenance chip".
- `AppListPanel.tsx` now says `removable` reads `AppStore.manifestIds`, not `source`.

**Merge-risk note on stale or missing manifest membership — checked, no change made.** Two claims, taken separately:

*Missing membership.* Every production `setCatalog` call passes `manifestIds` — `useAppManager.ts:146` (recompose), `:325` (refresh), `:464` (init) — and all three take it from the same `composeCatalog` return that produced `entries` and `sources`. There is no production path that writes `catalog` without `manifestIds`. The two-argument fallback in `appStoreImpl.setCatalog` exists only so existing tests keep compiling; it derives the ids from the `'manifest'`-sourced entries, which is the pre-PR behavior.

*Stale membership.* `recomposeCatalog` reads `manifestEntriesRef.current`, so an in-flight `refreshCatalog` leaves it pointing at the previous manifest — but `catalog`, `sources`, and `manifestIds` all come from that same snapshot, so they cannot disagree with each other. This is unchanged from before the PR.

The one state worth naming: `fetchManifest` returns `[]` on a failed fetch (`fetchManifest.ts:22`), so a failed manifest load yields empty `manifestIds` and a pinned app that is normally manifest-shipped does show Uninstall. That is correct for that session — with no manifest, `recomposeCatalog` has nothing to re-add the app from, so the uninstall sticks. `useAppManager.test.tsx` already exercises this: `obtainCatalogEntries` is mocked to `[]` and "removes an installed app from the workspace and catalog" passes.

**Title.** Renamed to `fix(app-manager): stop offering Uninstall for apps.json apps (#699)`.

PR #710, opened 2026-09-03
fix(toolbar): make the menu bar behave like a desktop menubar

### LINKED ISSUE #709: Toolbar menus need two clicks to switch and never switch on hover — no single source of truth for the open menu
## Problem

The toolbar menu bar does not behave like a standard desktop menubar.

- With one menu open (say **Data**), clicking **Help** does not open Help. The first click only closes Data; a second click is needed to open Help.
- Hovering across the other menu titles while a menu is open does nothing. In a desktop menubar (and in libraries such as PrimeReact's `NavigationMenu`) the open menu follows the pointer once any menu is open.
- There is no keyboard path between menus: ArrowLeft/ArrowRight do not move along the bar.

## Cause

Each top-level menu (`DataMenu`, `EditMenu`, `LayoutMenu`, `AnalysisMenu`, `ToolsMenu`, `AppMenu`, `HelpMenu`) owns its own `open` flag and renders a MUI `Popover`. A `Popover` is a modal with an invisible full-screen backdrop, so while one menu is open every click on the page — including one on another trigger — only hits the backdrop and closes the open menu. There is no shared state that knows which menu is open, so nothing can hand the open menu from one trigger to the next.

A related bug: a keydown inside a submenu (a `Popper` portal) bubbles through the React tree to the parent level's handler, so pressing ArrowDown in **Data › Import** moves focus back to the parent menu's first row.

## Expected

- One click opens a menu; while any menu is open, hovering another trigger or pressing ArrowLeft/ArrowRight moves the open menu there in one step.
- A single source of truth for which menu is open, so at most one is open at a time.
- Keyboard: ArrowDown/Enter open with focus on the first row, Home/End within a level, ArrowRight opens a submenu or moves to the next menu, ArrowLeft closes a submenu or moves to the previous menu, Escape closes and returns focus to the trigger, Tab or a click elsewhere closes.
- Hover switching is mouse-only, since a tap on touch devices fires hover right before its click.

## Steps to reproduce

1. Open the app and click **Data**.
2. Click **Help** — nothing opens; click it again and it opens.
3. With **Data** open, hover **Apps** — Apps does not open.

chrtannus:
Fixes #709.

## Summary

- **Root cause.** Each top-level menu kept its own `open` flag and rendered a MUI `Popover`, a modal with an invisible full-screen backdrop. While one menu was open, a click on any other trigger only hit that backdrop and closed the first menu, so every switch cost two clicks and hovering could never move the open menu.
- **`MenuBar`** (`src/features/ToolBar/MenuBar.tsx`, new) wraps the seven menus and is the single source of truth for which menu is open. Each menu takes its flag from `useMenuBarMenu(id)`; opening one closes whichever was open. The menu components change by one line each.
- **`DropdownMenu`** renders a non-modal `Popper` + `ClickAwayListener` instead of a `Popover`. Once a menu is open, hovering another trigger (mouse only — a tap fires hover right before its click) or pressing ArrowLeft/ArrowRight moves the open menu there.
- **Keyboard** follows the WAI-ARIA menubar pattern: ArrowDown/Enter/Space open with focus on the first row; Home/End within a level; ArrowRight opens a submenu or moves to the next menu; ArrowLeft closes a submenu or moves to the previous menu; Escape closes and returns focus to the trigger; Tab or a click elsewhere closes.
- **Submenu bug fixed.** A keydown inside a portaled submenu bubbled to the parent level, which moved focus onto its own first row on ArrowDown.
- Docs: `ToolBar.md` behaviour section; two lessons in `.serena/memories/lessons.md` (Portal content mounts one render after `open`; Popover's backdrop makes it unfit for sibling triggers).

## Notes

- The page underneath stays interactive while a menu is open, as in most web menubars (the modal backdrop is gone).
- The bar is `role="menubar"` (labelled "Main menu") and its triggers are `role="menuitem"`; a standalone `DropdownMenu` keeps the native button role. The logo sits beside the menubar, not inside it. (Review follow-up, second commit.)
- No dependency changes.
- **ArrowRight inside a submenu moves to the next menubar menu on purpose.** CodeRabbit asked to gate `onNavigateMenuBar(1)` on `topLevel`; not applied. The [WAI-ARIA APG menubar pattern](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/#keyboardinteraction) says ArrowRight on a menu item without a submenu closes the submenu chain, moves focus to the next menubar item and opens it (native macOS/Windows menubars do the same), so the handler is threaded down to every level; a row that owns a submenu stops the event first and opens the submenu instead. ArrowLeft is the mirror: it closes the current submenu and only at the top level moves to the previous menu. Gating on `topLevel` would make ArrowRight a no-op on submenu leaves, the one place the spec requires it to act. Discussed in [this comment](https://github.com/cytoscape/cytoscape-web/pull/710#issuecomment-5532440220).

## Test plan

- [x] New `MenuBar.spec.tsx` (10 tests: single-click switch, hover switch, touch and disabled-trigger guards, arrow keys across the bar, ArrowDown from a closed trigger, Escape focus return, focus-leave close, click-away close, parent-driven close) and a submenu ArrowDown regression test in `DropdownMenu.spec.tsx` — both failed before the fix.
- [x] `npm run test:checks:quiet` passes.
- [x] e2e (chromium): `undo-redo`, `onboarding-flow`, `local-file-import`, `dialog-dismiss`, `remote-app-load` — 16 passed.
- [x] Review follow-ups: ArrowUp from a pointer-opened menu enters at the last row; `menubar`/`menuitem` roles asserted in `MenuBar.spec.tsx`. Unit, lint and the same five e2e specs re-run green.
- [x] Second review pass (0693f6df): App menu imports lost in the `development` merge restored; hover switching is mouse-only (`pointerType === 'mouse'`, touch/pen/unset covered in `MenuBar.spec.tsx`); template rows honour the model's `disabled` flag (covered in `DropdownMenu.spec.tsx`). 91 toolbar unit tests, lint, typecheck and the toolbar e2e specs green.
- [x] Manual in the browser: click Data → hover Apps → hover Help → single click Data; ArrowRight/ArrowLeft across the bar skipping disabled menus; Data › Import submenu via ArrowRight/ArrowLeft; Escape.

## Summary by CodeRabbit

- **New Features**
  - Added coordinated toolbar menu behavior, allowing menus to switch smoothly between triggers with mouse or keyboard navigation.
  - Improved dropdown and submenu accessibility with Arrow, Home, End, and Escape key support.
  - Added focus management, including focus restoration when menus close and containment within open submenus.
  - Menus now close reliably when users click outside them.
  - Improved submenu keyboard navigation and support for skipping disabled menu items.

- **Documentation**
  - Updated toolbar documentation with menu behavior, navigation, focus handling, and submenu guidance.


chrtannus:
@coderabbitai Re **"Limit ArrowRight menubar switching to the top-level menu"** (`DropdownMenu.tsx`, the `ArrowRight` case in `MenuLevel`): not applied, on purpose.

The nested behaviour is the WAI-ARIA menubar pattern, not a leak of the top-level handler. From the [APG menubar keyboard interaction](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/#keyboardinteraction), for a menu opened from a menubar:

> **Right Arrow:** When focus is in a menu and on a menuitem that has a submenu, opens the submenu and places focus on its first item. Otherwise, closes the submenu and any parent menus, moves focus to the next item in the menubar, and, if focus is now on a menuitem with a submenu, opens the submenu.

So ArrowRight on a leaf inside **Data › Import** is supposed to leave the submenu chain and open **Apps** (Edit/Layout/Analysis/Tools being disabled with an empty workspace) — the same thing native menubars on macOS and Windows do. That is why `onNavigateMenuBar` is threaded down to every level, while the row handler for an item *with* a submenu stops the event first so ArrowRight there opens the submenu instead (`activateRow` + `stopPropagation`). ArrowLeft is the mirror: it closes the current submenu, and only at the top level moves to the previous menu.

Gating on `topLevel` would make ArrowRight a no-op on submenu leaves, which is the one place the spec says it must do something.

The other three findings from this pass are applied in 0693f6df: the App menu imports lost in the `development` merge, mouse-only hover switching (`pointerType === 'mouse'`), and template rows honouring the model's `disabled` flag.

PR #711, opened 2026-09-03
fix(workspace): mark networkModified at one choke point (#680)

### LINKED ISSUE #680: App API table writes leave networkModified false, so NDEx-backed networks skip the save
<details>
<summary>Original request (2026-08-24)</summary>

there is an issue where  @cw5/src/app-api/core/tableApi.ts does not set the network modified status to true, when adding columns

Do an audit of the networkmodified flag, figure out how we can refactor this to be properly set in both the api and in the main application

</details>

---

## Summary

`src/app-api/` never calls `setNetworkModified`. Table writes through the app API therefore leave `networkModified[networkId]` at `false`, and an NDEx-backed network with app-added columns is skipped by Save Workspace and shows a disabled "Save to NDEx" entry. The flag has no choke point: 27 in-app call sites set it by hand, and two `WorkspaceEditor` subscriptions cover the rest — but only for `currentNetworkId`, and only through the view model and visual style stores.

## Expected behavior

Any app API write that changes network data marks the target network modified, exactly as the equivalent in-app action does.

## Actual behavior

`tableApi.createColumn` adds the column and leaves the flag `false`. The user cannot save the network and Save Workspace silently omits it.

## Root cause

`grep -rn setNetworkModified src/app-api/` returns zero non-test hits. No store action sets the flag either — `TableStore.ts` never imports `WorkspaceStore` — so every mutation path has to remember to set it.

Two of the five mutating API modules get the flag by accident. `WorkspaceEditor.tsx:158` subscribes to the view model and `WorkspaceEditor.tsx:187` to the visual style store, flipping the flag when either changes. That covers `elementApi`, `visualStyleApi`, `viewportApi`, and `layoutApi` — but the selector is `state.getViewModel(currentNetworkId)`, while every app API method takes an explicit `networkId`.

`tableApi` writes only `TableStore`, which neither subscription watches. Nothing marks it, ever.

**Reproduced** with the real stores (throwaway test, not committed):

```ts
useTableStore.getState().add(networkId, createTable(networkId), createTable(networkId))
useWorkspaceStore.getState().setNetworkModified(networkId, false)

const res = tableApi.createColumn(networkId, 'node', 'newCol', 'string', 'x')
expect(res.success).toBe(true)                                                  // passes
expect(useTableStore.getState().tables[networkId].nodeTable.columns
  .map((c) => c.name)).toContain('newCol')                                      // passes
expect(useWorkspaceStore.getState().workspace
  .networkModified[networkId]).toBe(false)                                      // passes — the defect
```

## How the flag works today

`networkModified: Record<IdType, boolean | undefined>` lives on the workspace, not on the network (`Workspace.ts:16`, comment "Network is edited locally or not"). Three actions write it, all on `WorkspaceStore`:

| Action | Impl | Effect |
| --- | --- | --- |
| `setNetworkModified(id, isModified)` | `workspaceImpl.ts:153` | Sets the key to `true` or `false`. |
| `deleteNetworkModifiedStatus(id)` | `workspaceImpl.ts:170` | Removes the key. |
| `deleteAllNetworkModifiedStatuses()` | `workspaceImpl.ts:185` | Empties the record. |

Readers all use `?? false` (`workspaceApi.ts:111`, `NetworkPropertyPanel.tsx:143`) or `=== true` (`useSaveWorkspaceToNDEx.ts:139`), so absent and `false` are equivalent. The two clear mechanisms are interchangeable.

Every write persists. `WorkspaceStore`'s `persist` middleware (`WorkspaceStore.ts:61-102`) writes the whole workspace row to IndexedDB on any workspace change, and `crossTabHydration.ts:141` hydrates that row into other tabs, so the flag is durable and cross-tab. `useLoadWorkspace.ts:112` resets it to `{}` when a remote workspace is loaded. The schema validator requires the field (`validator.ts:277`).

There is no `network:modified` event on the app API event bus. It dispatches `network:created`, `network:deleted`, `network:switched`, `network:changed`, `selection:changed`, `style:changed`, and `data:changed`, but none of them carries the flag, so an app tracking save state has to poll `workspaceApi.getNetworkSummary`.

## Everything that sets the flag

**Implicit — two subscriptions, both scoped to `currentNetworkId`**

| Site | Watches | Notes |
| --- | --- | --- |
| `WorkspaceEditor.tsx:158-179` | `getViewModel(currentNetworkId)` | Diffs the view model with `selectedNodes`/`selectedEdges` omitted, so selection does not count as a modification. Covers node/edge create and delete, node moves, layouts. |
| `WorkspaceEditor.tsx:187-203` | whole `VisualStyleStore` | Deep-equals `visualStyles[currentNetworkId]`. Covers every Vizmapper and `visualStyleApi` write. |

Both are registered in the component body rather than in a `useEffect` — see the leak below.

**Explicit `true` — 27 live sites**

| File:line | Operation |
| --- | --- |
| `TableToolbar.tsx:194` | Sort column ascending (display config) |
| `TableToolbar.tsx:206` | Sort column descending (display config) |
| `TableToolbar.tsx:215` | Duplicate column (table) |
| `TableToolbar.tsx:242` | Duplicate column (display config) |
| `TableToolbar.tsx:275` | Rename column (table) |
| `TableToolbar.tsx:294` | Rename column (display config) |
| `TableToolbar.tsx:329` | Delete column (table) |
| `TableToolbar.tsx:346` | Delete column (display config) |
| `TableToolbar.tsx:386` | Apply value to all rows |
| `TableToolbar.tsx:416` | Apply value to selected rows |
| `TableToolbar.tsx:482` | Insert new column (table) |
| `TableToolbar.tsx:502` | Insert new column (display config) |
| `cellEditHandler.ts:75` | Cell edit — single value |
| `cellEditHandler.ts:95` | Cell edit — list value |
| `cellEditHandler.ts:117` | Cell edit — multi-row fill |
| `contextMenuActions.ts:86` | Table context menu — delete column |
| `contextMenuActions.ts:146` | Table context menu — rename column |
| `columnHandlers.ts:97` | Column move (display config) |
| `columnHandlers.ts:187` | Column resize (display config) |
| `pasteHandler.ts:87` | Paste into table |
| `useListEditor.ts:80` | List-cell editor commit |
| `StyleManager.tsx:101` | Apply / rename / delete a named visual style |
| `Checkbox.tsx:24` | Vizmapper boolean visual property (default) |
| `Checkbox.tsx:58` | Vizmapper boolean visual property (bypass) |
| `NetworkPropertyEditor.tsx:211` | Edit network name / description |
| `TableColumnAppendForm.tsx:192` | Join a table file into the network |
| `updateNetwork.ts:140` | Service-app `updateNetwork` result applied |

(`TableBrowser.tsx:193` is a 28th, commented out.)

**Cleared — 4 sites**

| File:line | When |
| --- | --- |
| `useRegisterNetwork.ts:142` | Network first registered — starts clean |
| `layoutCompletion.ts:75` | Automatic initial layout finished and nothing else changed |
| `SaveToNDExMenuItem.tsx:162` | Single network saved to NDEx |
| `useSaveCurrentNetworkToNDEx.ts:165` | Current network saved to NDEx |

**Removed — 2 sites** (`deleteNetworkModifiedStatus`): `useSaveWorkspaceToNDEx.ts:152` after the network uploads, `deleteNetworkOrchestrator.ts:45` on network delete (`:86` clears all on workspace reset).

**Sets it nowhere** — these rely entirely on the view model subscription: `useCreateNode.ts`, `useCreateEdge.ts`, `useDeleteNodes.ts`, `useDeleteEdges.ts`, `useUndoStack.tsx`, `LayoutMenu/index.tsx`, `ApplyLayoutButton.tsx`, `Scaling.tsx`, `MergeDialog.tsx`, `SubNetworkPanel.tsx`, and all 5 mutating app API modules.

## What the unset flag costs

| Consumer | Behavior when the flag is false |
| --- | --- |
| `useSaveWorkspaceToNDEx.ts:139` | NDEx-backed network is skipped. The app's column never reaches NDEx. Local networks are saved as a copy regardless (`:123`), so the loss is NDEx-only. |
| `networkSaveStatus.ts:28` | `getSaveButtonState` returns `action: 'none'`, tooltip "Up to date with NDEx". |
| `networkSaveStatus.ts` (`getSaveMenuItemState`) | Save entry is `disabled: true`, hint "No unsaved changes". The user cannot save manually either. |
| `workspaceApi.ts:111,141` | `isModified` reported to other apps is wrong. |
| `NetworkPropertyPanel.tsx:143` | Unsaved-changes badge absent. |

## Two more defects in the same mechanism

**The subscriptions leak.** `useViewModelStore.subscribe` (`WorkspaceEditor.tsx:158`) and `useVisualStyleStore.subscribe` (`WorkspaceEditor.tsx:187`) sit in the component body, not in a `useEffect`. Each render registers another listener and the returned unsubscribe is discarded, so the listener count grows for the life of the session. Every listener closes over the render-time `workspace` object, so the `isCurrentNetworkUnmodified` guard reads a stale flag. The writes are idempotent, which is why this has not shown up as a visible bug.

**`tableApi` records no undo.** It does not import `corePostEdit` (`elementApi.ts` and `layoutApi.ts` do). Table edits made through the API are not undoable. Same missing choke point, same fix location.

## Is the current model compatible with the app API

The data model is fine. `Record<IdType, boolean>` on the workspace, persisted, read through `?? false` — that answers the only question anyone asks it ("does this network have unsaved work?") and it survives a reload, which it must, or unsaved work looks saved. Keep it.

The **write path** is what does not carry over. Three specific mismatches:

**1. The write path is keyed on the displayed network; the API is keyed on an argument.** Both subscriptions select on `currentNetworkId`, and 22 of the 27 explicit sites pass `currentNetworkId` too. That held while every edit came from a component rendering the network on screen. It does not hold now: `NetworkStore.networks` is a `Map` with no eviction on network switch (only `deleteNetworkOrchestrator.ts:40` and `:79` remove entries), so networks accumulate as the user visits them, and HierarchyViewer sub-networks and merge results are resident too. `tableApi.createColumn(residentNonCurrentId, …)` succeeds and nothing marks it — the same `currentNetworkId`-instead-of-`networkId` mistake #592 reports in the persist middleware.

**2. There is no store-level choke point, so every new module has to remember.** No store action writes the flag. Correctness rests on 27 hand-written calls plus two subscriptions that happen to catch four more paths. The app API added five mutating modules and none of them remembered. The next module will not either.

The app API's own event bus already solves the scoping half of this correctly, in the same repo. `initEventBus.ts:233` subscribes to `TableStore.tables` and iterates **every** networkId, diffing each node and edge table to emit `data:changed` with added and removed columns. That is exactly the table-change detection the modified flag is missing, written networkId-first rather than `currentNetworkId`-first.

**3. Apps can read the flag but not write it.** `workspaceApi` exposes `isModified` read-only. An app that uploads the network itself with its own NDEx credentials cannot clear the flag, so the network stays marked unsaved forever. An app making a change the user should not be asked to save cannot suppress the mark. And with no `network:modified` event, an app tracking save state has to poll.

A reimplementation has to preserve two subtleties that are load-bearing today:

- Selection is not a modification (`WorkspaceEditor.tsx:167-170` omits `selectedNodes`/`selectedEdges` before diffing).
- The automatic initial layout is not a modification, but a user edit racing it is (`layoutCompletion.ts:63-76` snapshots the flag before writing positions).

## Proposed approach

Keep the model; replace the write path with one choke point keyed on the mutated `networkId`.

Add `markNetworkModified(networkId)` to `src/app-api/core/undo.ts`, beside `corePostEdit`, and call it from the top of `corePostEdit` — before the `UNDO_STACK_SIZE > 0` branch, so a deployment with `undoStackSize: 0` still marks. Mirror the same call in `useUndoStack.postEdit`, which covers the 42 in-app `postEdit` sites. Then delete the two `WorkspaceEditor` subscriptions and the explicit calls they duplicate.

`corePostEdit` is the right hook because every mutation already routes through it with an explicit `networkId`, and because load and undo stay clean without a suppression flag: `useRegisterNetwork` and `loadCyNetwork` post no edits, and undo/redo replays store actions without calling `postEdit`.

Three modules record no undo and need explicit `markNetworkModified` calls: `tableApi` (8 writes), `visualStyleApi` (9 writes), `viewportApi.updateNodePositions`. Adding `corePostEdit` to `tableApi` is worth doing in the same pass — table edits made through the API are currently not undoable.

| Rejected | Why not |
| --- | --- |
| Add `setNetworkModified` to each app API mutator only | Fixes the API and leaves the in-app drift and the leaked subscriptions in place. |
| Set the flag inside each mutating store action | Catches load, import, and undo replay as user edits; needs a suppression flag threaded through every load path. |
| Fix the subscriptions instead — iterate all networkIds and add a `TableStore` one, modelled on `initEventBus.ts:233` | No call sites to change and impossible for a new module to miss, but every load, import, and cross-tab hydration writes to those stores, so each subscription needs an `isHydrating()` guard (`hydrationContext.ts:15`) and undo replay still needs suppressing. A spurious `data:changed` is harmless; a spurious modified flag is not. Worth revisiting if the choke point proves leaky. |
| Derive the flag by diffing against the last-saved network | Correct by construction, but needs a saved-state snapshot per resident network. Out of proportion to the defect. |

## Scope of change

- `src/app-api/core/undo.ts` — add `markNetworkModified`; call it from `corePostEdit`
- `src/app-api/core/tableApi.ts` — 8 write methods mark modified; also add `corePostEdit`
- `src/app-api/core/visualStyleApi.ts` — 9 write methods mark modified
- `src/app-api/core/viewportApi.ts` — `updateNodePositions` marks modified
- `src/data/hooks/useUndoStack.tsx` — `postEdit` marks its target network
- `src/features/Workspace/WorkspaceEditor.tsx:158-203` — delete both subscriptions
- `src/features/TableBrowser/components/TableToolbar.tsx` — drop the 12 now-redundant calls
- `src/features/TableBrowser/utils/{columnHandlers,contextMenuActions,cellEditHandler,pasteHandler}.ts` — same, and drop `setNetworkModified` from their args interfaces
- `src/features/TableBrowser/{TableBrowser.tsx,components/TableContextMenu.tsx}`, `src/features/TableBrowser/hooks/useListEditor.ts` — stop threading the action through props
- `src/features/Vizmapper/{StyleManager/StyleManager.tsx,VisualPropertyRender/Checkbox.tsx}` — same
- `src/features/SummaryPanel/NetworkPropertyEditor.tsx:211` — keep; summary edits post no undo
- `src/features/TableDataLoader/components/JoinTableToNetwork/TableColumnAppendForm.tsx:192` — keep or route through `postEdit`
- `src/features/ServiceApps/resultHandler/updateNetwork.ts:140` — verify still needed
- `src/app-api/core/tableApi.test.ts` — mock `WorkspaceStore`; assert the flag on all 8 writes
- `src/features/Workspace/layoutCompletion.test.ts` — the snapshot-before-write ordering comment names the deleted subscription; update it
- `docs/specifications/MULTIPLE_VISUAL_STYLES.md:201` — restates "all operations mark the network modified"; confirm still true
- `src/features/Workspace/Workspace_docs/WorkspaceEditor.md:56` — documents the deleted subscription

## Open questions

- Should display-only changes mark the network modified? Column sort (`TableToolbar.tsx:194,206`), move (`columnHandlers.ts:97`), and resize (`columnHandlers.ts:187`) all set it today, so sorting a table makes a clean network unsaved. `tableDisplayConfiguration` does ship in CX2, so this may be intentional.
- Should undoing back to the original state clear the flag? It does not today.
- Does the app API need a write side — `setModified(networkId, boolean)` on `workspaceApi`, plus a `network:modified` event? Apps can already observe table and style writes through `data:changed` and `style:changed`, but not the flag itself, and an app that saves the network with its own credentials cannot clear it. Out of scope here unless wanted.
- Keep both clear mechanisms (`setNetworkModified(id, false)` and `deleteNetworkModifiedStatus(id)`)? Both read back as `false`.

## Verification

1. `src/app-api/core/tableApi.test.ts` — one case per write method asserting `networkModified[networkId] === true` after the call, plus a case asserting a failed write does not set it.
2. `src/app-api/core/elementApi.test.ts` and `visualStyleApi.test.ts` — assert the flag is set for a `networkId` that is **not** `currentNetworkId`. That is the case the deleted subscription could never cover.
3. `WorkspaceEditor` regression — render twice and assert the store's listener count does not grow.
4. `src/features/Workspace/layoutCompletion.test.ts` — must still pass: the initial layout leaves the flag clean, a user edit during layout survives it.
5. `npm run test:checks:quiet`.
6. E2E: `npm run e2e:spec -- network-property-menu` — `network-property-menu.spec.ts:262` ("overflow button is badged only while the network has unsaved changes") already pins the badge to the flag. Add a case that drives `window.CyWebApi` table writes and asserts the badge appears.

## Related

- #592 — Data Persistence and Architecture Robustness. Same `currentNetworkId`-instead-of-`networkId` mistake, in the persist middleware.
- #553 — Deleting columns through the API doesn't seem to work (closed).

d2fong:
Fixes #680.

## The defect

App API writes left `networkModified[networkId]` at `false`. An NDEx-backed network with app-added columns was therefore skipped by Save Workspace and showed a disabled "Save to NDEx" entry — `tableApi.createColumn` added the column and the user could not save it.

The flag had no choke point. 27 hand-written `setNetworkModified` calls, plus two `WorkspaceEditor` store subscriptions that covered the rest — but those selected on `currentNetworkId`, while every app API method takes an explicit `networkId` and non-current networks stay resident in the stores (`NetworkStore` evicts only on delete).

## The fix

`markNetworkModified(networkId)` in `src/app-api/core/undo.ts`, called from `corePostEdit` and from `useUndoStack.postEdit`. Both call it **before** the stack write, so a deployment with `undoStackSize: 0` still marks. Undo and redo replay mark too: replaying an edit after a save leaves unsaved work, which the deleted subscriptions used to catch.

Both `WorkspaceEditor` subscriptions are deleted. They sat in the component body rather than an effect, so every render added a listener and discarded its unsubscribe, and each listener read a stale `workspace` from its render closure.

Paths that record no undo entry mark explicitly:

| Path | Why no undo entry |
| --- | --- |
| `tableApi.createColumn` | No `CREATE_COLUMN` command exists (in-app "Insert new column" is not undoable either) |
| `tableApi.importTableFromTsv` | Creates columns **and** writes cells; a cell-only entry would restore values and leave the new columns behind |
| `visualStyleApi` — all 9 writes | No style undo commands are wired to the API |
| `viewportApi.updateNodePositions` | `MOVE_NODES` is per-node; a batch reposition has no matching command |

Six `tableApi` writes gained `corePostEdit` against the existing command contracts, so table edits made through the API are undoable for the first time: `setValue`, `setValues`, `editRows`, `applyValueToElements`, `deleteColumn`, `renameColumn`.

## Explicit calls removed

Only where a `postEdit` on the same action, with the same network, provably covers it:

- `TableToolbar` — 6 of 12 (rename column ×2, delete column ×2, apply-to-column, apply-to-selected)
- `contextMenuActions` — both; `cellEditHandler` — all 3; `pasteHandler`, `useListEditor` — 1 each
- `StyleManager` — 1 of 7 (`handleSwitch`)
- `NetworkPropertyEditor` — now passes its `networkId` to `postEdit` instead

`setNetworkModified` is no longer threaded through `TableContextMenu`, `cellEditHandler`, `pasteHandler`, `useListEditor` or `contextMenuActions`.

## Deviation from the ticket

The ticket listed all of `TableToolbar`, `Checkbox.tsx` and `StyleManager` for removal. Six of `TableToolbar`'s twelve calls have no `postEdit` behind them (sort asc/desc, duplicate ×2, insert column ×2), and neither do `columnHandlers` (column move/resize), `Checkbox.tsx` (writes `visualStyleOptions`, not the visual style), or `StyleManager`'s duplicate/rename/delete. Dropping those would have lost the mark outright. They stay, each with a comment naming the reason. `TableColumnAppendForm` and the service-app `updateNetwork` stay for the same reason.

## Two behavior notes

**`tableImpl.applyValueToElements` branches on `elementIds != null`.** Omitting the argument applies to every row; passing `[]` applies to none. The undo snapshot and the mark follow that split, and a no-op write (`[]`, no cell edits, no rows) records nothing and marks nothing.

**HierarchyViewer sub-network layouts no longer mark the parent hierarchy modified.** The view model subscription attributed them to `currentNetworkId`, which produced a spurious unsaved badge.

## Verification

`npm run test:checks:quiet` — green, 4046 unit tests.

New coverage:

1. `tableApi.test.ts` — the flag on all 8 write methods; failure and no-op cases assert it stays clear; undo params pinned against the `useUndoStack` handler contracts (`SET_CELL_VALUE` params, `DELETE_COLUMN`'s pre-delete table and `params[3].id`, `RENAME_COLUMN`'s swapped names).
2. `elementApi.test.ts`, `visualStyleApi.test.ts`, `viewportApi.test.ts` — the flag is set for a `networkId` that is **not** `currentNetworkId`. That is the case the deleted subscriptions could never cover.
3. `undo.test.ts` — `corePostEdit` marks, and still marks at `undoStackSize: 0`.
4. `WorkspaceEditor.spec.ts` — source guard against re-adding a store subscription or a direct `setNetworkModified(id, true)` to the component. A source assertion rather than a render test: rendering the real `WorkspaceEditor` needs a router, the app config context, a hydrated IndexedDB and several lazy children, and the invariant is fully visible in the source.
5. `layoutCompletion.test.ts` — the comment naming the deleted subscription is updated; two cases added for the current behavior, where the position write marks nothing.

E2E (chromium): `network-property-menu` 11 passed — including a new case driving `window.CyWebApi.table.createColumn` and asserting the unsaved badge appears. `undo-redo`, `app-api-column-persistence`, `table-browser` also passed.

## Left at current behavior

Three of the ticket's open questions, unchanged here:

- Display-only changes (column sort, move, resize) still mark the network.
- Undoing back to the original state does not clear the flag.
- The app API has no write side (`workspaceApi.setModified`, a `network:modified` event) — the ticket scoped it out.

## Docs

- `src/features/Workspace/Workspace_docs/WorkspaceEditor.md` — the Modification Tracking and Modification Detection Flow sections
- `docs/specifications/MULTIPLE_VISUAL_STYLES.md` §5 — which style operations mark, and that "Save Style to Library" does not
- `.serena/memories/lessons.md` — the choke point, the `applyValueToElements` trap, and that `vi.clearAllMocks()` does not reset mock implementations

## Summary by CodeRabbit

* **New Features**
  * Network edits now consistently trigger the unsaved-changes indicator, including table, visual-style, viewport, undo, and redo operations.
  * Table edits now support undo and redo with accurate restoration of affected values.
  * No-op or unsuccessful operations no longer mark networks as modified.
* **Bug Fixes**
  * Edits correctly mark the targeted network, even when it is not currently active.
  * Networks are marked modified even when undo history is disabled.
* **Tests**
  * Added coverage for network modification tracking, table undo/redo behavior, and unsaved-changes indicators.


d2fong:
@coderabbitai review

PR #720, opened 2026-09-10
fix(AppManager): name why an app failed to load, and hide Retry when it cannot help (#719)

### LINKED ISSUE #719: AppManager: failed app loads show no reason, and Retry cannot fix most of them
<details>
<summary>Original request (2026-09-10)</summary>

The bug you found
chrisapp (catalog) ≠ chrisApp (bundle). App is mis-packaged. Not your fault, not fixable by you.

Retry can never fix it. The message was the only thing missing.

The fix, in 4 pieces
1. Store the reason
Add loadErrors: Record<string, AppLoadFailure> next to loadStates.
New action setLoadFailed(id, failure) sets both.
Don't widen AppLoadState — too many === 'failed' call sites.

2. Make loadRemoteApp return the reason
Today it returns undefined for three different failures. Return {ok, failure} instead.
2 callers to update.

3. Show it twice

Toast when it happens (dialog is usually closed)
Red caption on the row, persistent (not a tooltip)
4. Make Retry honest
Of 8 failure modes, Retry helps with one.

Failure	Retry?
mount-failed	✅ yes
fetch-failed	⚠️ needs page reload
other 6	❌ no button, just the reason
Two things not to skip
Don't lowercase the id to "fix" it. mountedApps is keyed by bundle id, deactivateApp looks it up by catalog id. Accept a mismatch → unmount never fires → panels leak.

Startup path fails silently today. Broken app fails on every page load, no event. Collapse those into one toast: "2 apps failed to load."

Out of scope for v1
Details dialog · persisting failures to IndexedDB · error-reporting hook · cache-busting retry

</details>

---

## Summary

A failed app load shows the user a bare `failed` chip and nothing else. The reason exists
only as a `logApp.warn` line, visible with debug mode on. Store the reason in `AppStore`,
render it on the row and in a toast, and show the Retry button only for the one failure
mode it can fix.

## Expected behavior

A failed row names the cause and what to do about it. Retry appears only when retrying
can succeed.

## Actual behavior

The row shows `label="failed"` (`src/features/AppManager/AppListPanel.tsx:213`) and a
Retry button (`AppListPanel.tsx:67`). No cause reaches the UI. `AppLoadState` is
`'unloaded' | 'loading' | 'loaded' | 'failed'`
(`src/models/AppModel/AppLoadState.ts:5`) and carries no error payload, so nothing stores
one.

Reproduced with a mis-packaged app from an allowed origin. With debug mode on:

```
app:warn [loadRemoteApp]: Remote app id mismatch. Expected "chrisapp", received "chrisApp" from https://apps-stage.cytoscape.org/web/chrisapp/0.2.0/remoteEntry.js +0ms
```

The catalog id and the bundle's own `CyApp.id` differ by one character's case. Without the
debug logger there is no way to learn this, and Retry cannot fix it.

## Test fixtures

Two causes have hosted fixtures. The other three are unit-test-only — no public bundle
reproduces them.

**id-mismatch (hosted)**
- URL: `https://apps-stage.cytoscape.org/web/chrisapp/0.2.0/remoteEntry.js` (HTTP 200)
- Catalog id `chrisapp`; bundle declares `chrisApp`.
- Expected: row states both ids and that the app is mis-packaged. No Retry button.

**fetch-failed (hosted)**
- URL: `https://apps-stage.cytoscape.org/web/does-not-exist/remoteEntry.js`
- Allowed origin, missing path — the request 404s.
- Expected: row names the URL and the HTTP status. Affordance is "reload to retry", not Retry.

**origin-blocked (hosted)**
- Any `https:` URL outside `appInstallAllowedOrigins`, set through Manifest Source.
- Expected: row names the rejected origin. No Retry button. A toast already fires here
  (`useAppManager.ts:235`) — this is the one path with a usable message today.

**no-app-config (unit test only)**
- A remote whose `./AppConfig` module has no `default` export.
- Expected: row says the app is mis-packaged. No Retry button.

**mount-failed (unit test only)**
- An app whose `mount()` throws.
- Expected: row shows the thrown message. **Retry button present** — this is the only
  retryable cause.

## Root cause

Two separate discards.

`loadRemoteApp` distinguishes three failures internally and returns `undefined` for all
of them: no default export (`loadRemoteApp.ts:29`), id mismatch (`loadRemoteApp.ts:36`),
and any thrown error (`loadRemoteApp.ts:46`). Its signature is
`Promise<CyApp | undefined>` (`loadRemoteApp.ts:20`), so the caller cannot recover which
one happened.

`AppStore` then has nowhere to put a reason. Seven call sites write
`setLoadState(id, 'failed')` with no second argument — `useAppManager.ts:243`, `:259`,
`:273`, `:293`, `:536`, `:578`, `:588` — and `setLoadState` only writes the string
(`AppStore.ts:299`, `appStoreImpl.ts:369`).

The startup auto-load path discards the reason a third time: it wraps the failure in a
synthetic `new Error()` (message: `Failed to load remote app "<id>"`) at `useAppManager.ts:561`
before `Promise.allSettled` collects it. That path also raises no toast at all, so a
mis-packaged active app fails silently on every page load.

## Retry is wrong for four of five causes

| Cause | Deterministic? | Retry outcome today |
| --- | --- | --- |
| `id-mismatch` | Yes | Fails identically, forever |
| `no-app-config` | Yes | Fails identically, forever |
| `origin-blocked` | Yes, until config changes | Fails identically |
| `fetch-failed` | No | Fails identically — see below |
| `mount-failed` | No | Can succeed |

`retryApp` is `await activateApp(id)` and nothing else (`useAppManager.ts:311`). For
`fetch-failed` it cannot succeed even when the network recovers: `ensureRemoteRegistered`
returns early when the scope is already bound to the same URL
(`src/features/AppManager/ExternalComponent.tsx:56`), and the federation runtime memoizes
the entry load in `globalLoading[name + entry]`
(`node_modules/@module-federation/runtime-core/dist/utils/load.js:134`), returning that
entry at `:198` without ever deleting the key on rejection. The rejected promise is
cached for the life of the tab. Only a page reload clears it.

## Scope of change

| File | Change |
| --- | --- |
| `src/models/AppModel/AppLoadFailure.ts` | New. Discriminated union of the five causes. |
| `src/models/AppModel/AppLoadState.ts` | Unchanged — see Proposed approach. |
| `src/models/StoreModel/AppStoreModel.ts` | Add `loadErrors: Record<string, AppLoadFailure>` beside `loadStates` (`:32`); add `setLoadFailed` to `AppAction`. |
| `src/models/StoreModel/impl/appStoreImpl.ts` | Add `setLoadFailed`; clear `loadErrors[id]` in `setLoadState` (`:369`) and in `removeApp` (`:399`, which already prunes `loadStates`). |
| `src/data/hooks/stores/AppStore.ts` | Wire `setLoadFailed` (beside `setLoadState`, `:299`). |
| `src/features/AppManager/loader/loadRemoteApp.ts` | Return `{ ok: true; app } \| { ok: false; failure }` instead of `CyApp \| undefined`. |
| `src/data/hooks/stores/useAppManager.ts` | Update both `loadRemoteApp` callers (`:271`, `:559`); convert 7 `setLoadState(id, 'failed')` sites to `setLoadFailed`; stop wrapping the failure in a synthetic Error at `:561`; add the aggregated startup toast. |
| `src/features/AppManager/AppListPanel.tsx` | Render the reason as caption text on the row; make `getAction` (`:60`) return `'retry'` only for `mount-failed`. |
| `src/features/AppManager/AppSettingsDialog.tsx` | Replace the raw `err.message` in the install-from-URL helper text (`:113`) with mapped text — pasting a `remoteEntry.js` URL there currently surfaces a raw `SyntaxError`. |
| `src/features/AppManager/loader/loadRemoteApp.test.ts` | 5 of 6 tests assert `undefined` (`:48`, `:66`, `:79`, `:92`, `:105`) — update to the result shape. |
| `src/features/AppManager/AppListPanel.spec.tsx` | Add cases for the reason text and for Retry's absence. |
| `src/data/hooks/stores/AppStore.spec.ts` | Add `setLoadFailed` and the clear-on-transition cases. |
| `src/features/AppManager/AppManager_docs/AppManager.md` | Document the failure codes under Status Management. |

## Proposed approach

Add `loadErrors` as a sibling record rather than widening `AppLoadState` into a
discriminated union. `AppListPanel` compares against the bare strings in four places
(`:67`, `:75`, `:138`, `:213`) and `useAppManager` writes them in seven; a union rewrites
all of them for no gain. Keep `loadErrors` session-local, matching `loadStates`.

Make `AppLoadFailure` carry data, not prose — `{ code: 'id-mismatch'; url; expected; received }`
— so the store holds no user-facing English and `getAction` can branch on `code` rather
than pattern-match a sentence.

Surface it twice: a toast at the moment of failure, because activation also happens at
startup and from `?installApp` intents when the App Manager is closed, and persistent
caption text on the row. Not a tooltip — undiscoverable, and unusable on touch. Collapse
startup failures into one message ("2 apps failed to load — open App Manager for
details"); `SnackbarMessageList` shows messages one at a time with a 300ms gap, so N
broken apps would otherwise queue N toasts.

For `id-mismatch` and `no-app-config` the user caused nothing and can fix nothing. The
message must say the app is mis-packaged and carry both ids verbatim so they can report it
to the publisher.

### Why the strict id check stays

Relaxing `remoteApp.id !== id` (`loadRemoteApp.ts:32`) to a case-insensitive compare looks
like the cheaper fix. It is not: the two ids are used as different keys. `mountApp` and
`unmountApp` key `mountedApps` by the bundle's `cyApp.id`
(`src/data/hooks/stores/appLifecycle.ts:37`, `:52`, `:74`, `:77`), while `deactivateApp`
tests `mountedApps.current.has(id)` with the catalog id (`useAppManager.ts:305`). The
catalog id is also the federation scope key, the `appRegistry` key, and the `loadStates`
key. Accept a mismatched pair and mount registers under `chrisApp` while deactivate looks
for `chrisapp`, so `unmountApp` never runs and the app's panels and menu items are never
cleaned up. The strict check is what keeps the two identities collapsed; today nothing can
diverge because it rejects the load. The packaging error stays an error.

## Out of scope

- A details dialog behind the chip — the row text is enough.
- Persisting failures to IndexedDB.
- Routing failures to `errorReportEndpoint`.
- Cache-busting retry. Clearing `globalLoading` to make in-tab retry work after a network
  failure reaches into federation-runtime internals and needs its own decision.

## Open questions

- Wording owner for the five user-facing messages.
- `fetch-failed`: relabel the button "Reload to retry" and trigger a reload, or show no
  button and say "reload the page to retry"?

## Verification

- `npx vitest run loadRemoteApp` — each failure returns its own `code`.
- `npx vitest run AppStore` — `setLoadFailed` sets both records; `setLoadState('loading')`
  and `removeApp` clear `loadErrors[id]`.
- `npx vitest run AppListPanel` — reason text renders; Retry present for `mount-failed`,
  absent for the other four.
- Manual, against the hosted fixtures above: install `chrisapp` from
  `apps-stage.cytoscape.org`, enable it, confirm the row names both ids and shows no Retry
  button, with debug mode **off**.

d2fong:
Closes #719

A failed app showed a bare `failed` chip and a Retry button. The reason existed
only as a `logApp.warn` line, and Retry could fix one of the five causes.

Reproduction: catalog id `chrisapp`, bundle id `chrisApp`. The app is
mis-packaged; nothing the user does fixes it, and only the debug logger said so.

## Changes

| File | Change |
| --- | --- |
| `src/models/AppModel/AppLoadFailure.ts` | New. Five-code discriminated union: `origin-blocked`, `fetch-failed`, `no-app-config`, `id-mismatch`, `mount-failed`. Carries identifiers, not prose. |
| `src/models/AppModel/impl/appLoadFailureMessage.ts` | New. Maps a code to user-facing text. |
| `src/features/AppManager/loader/loadRemoteApp.ts` | Returns `{ ok, app } \| { ok: false, failure }` instead of `CyApp \| undefined`, which flattened three causes into one value. |
| `src/models/StoreModel/AppStoreModel.ts`, `impl/appStoreImpl.ts`, `src/data/hooks/stores/AppStore.ts` | `loadErrors` beside `loadStates`; `setLoadFailed(id, failure)` writes both. `setLoadState` clears the reason, `remove` prunes it. |
| `src/data/hooks/stores/useAppManager.ts` | Seven `setLoadState(id, 'failed')` sites now record a reason. Startup path stops wrapping the failure in a synthetic Error and raises one aggregated toast. |
| `src/features/AppManager/AppListPanel.tsx` | Reason as persistent caption text on the row; Retry only for `mount-failed`. |
| `src/features/AppManager/AppSettingsDialog.tsx` | Install-from-URL no longer surfaces a raw `SyntaxError` when a `remoteEntry.js` URL is pasted where the manifest URL belongs. |

`AppLoadState` is unchanged. A union would rewrite eleven call sites for no gain.

## Retry is honest now

`retryApp` is `activateApp` and nothing else, so it helps with one cause:

| Cause | Deterministic? | Retry? |
| --- | --- | --- |
| `id-mismatch` | Yes | No — mis-packaged |
| `no-app-config` | Yes | No — mis-packaged |
| `origin-blocked` | Until config changes | No |
| `fetch-failed` | No | No — reload the page |
| `mount-failed` | No | **Yes** |

`fetch-failed` cannot succeed in-tab even when the network recovers:
`ensureRemoteRegistered` returns early when the scope is already bound to the
same URL, and the federation runtime memoizes the entry load in
`globalLoading[name + entry]` without deleting the key on rejection. The
rejected promise is cached for the life of the tab.

## Two decisions

**The open question on `fetch-failed`** — no button, and the caption says to
reload. A "Reload to retry" button that reloads the whole tab is a page-level
action wearing a row-level control's clothes, and the ticket's scope table
already called for `'retry'` only on `mount-failed`.

**The message mapper lives in the model layer**, not in the AppManager feature.
`src/data/layering.test.ts` records the `data → features` baseline and says
shrink it, never extend it; `useAppManager` needs the mapper for the toast. It
is pure TypeScript — no React, no Zustand — and the store still holds no
English.

## Why the strict id check stays

Relaxing `remoteApp.id !== id` to a case-insensitive compare looks cheaper. It
is not: the two ids key different maps. `mountApp`/`unmountApp` key
`mountedApps` by the bundle's `cyApp.id`, while `deactivateApp` tests
`mountedApps.current.has(id)` with the catalog id. Accept a mismatched pair and
mount registers under `chrisApp` while deactivate looks for `chrisapp`, so
`unmountApp` never runs and the app's panels and menu items leak.

## Verification

- `npm run lint` clean (tsc + oxlint).
- Unit suite: 4261 pass. Two runs each had one unrelated timeout —
  `ColorPalettePicker.spec.tsx`, then `ListValueEditorDialog.spec.tsx` — and
  both pass alone; different file each time, so it is worker contention.
- `npx vitest run loadRemoteApp AppStore AppListPanel useAppManager appLoadFailureMessage` — 105 pass.
- `remote-app-load` e2e: 4 passed, run with `E2E_DEV=1` against a running dev
  server rather than stopping it. That path is the flakier one, so CI's own run
  is the real signal.

## Not verified

- The two hosted fixtures in the ticket (`chrisapp` id-mismatch,
  `does-not-exist` 404) were not driven manually in a browser. Unit tests cover
  both codes from `loadRemoteApp` through to the rendered caption.
- Wording for the five user-facing messages is unreviewed — that open question
  in the ticket is still open.

## Out of scope (per the ticket)

Details dialog behind the chip · persisting failures to IndexedDB · routing to
`errorReportEndpoint` · cache-busting retry.

## Summary by CodeRabbit

* **New Features**
  * App loading failures now show specific, actionable reasons.
  * Failed apps display persistent error messages in the app list.
  * Retry is available for recoverable mount failures and hidden for non-retryable failures.
  * Startup failures are summarized in a single notification.
  * Installing an app from a URL now provides clearer HTTP, manifest, and connection error messages.
  * Refreshing an app with a corrected bundle URL clears its previous failure state.

* **Documentation**
  * Added documentation for load failure types and retry behavior.


d2fong:
> Restrict `setLoadState` to non-failed states.

Applied in 0511c93e. Verified first: every caller passes `'loading'` or
`'loaded'`, so the narrowing costs nothing at runtime and the only dynamic
pass-through (`AppStore.ts:303`, forwarding its own parameter) narrows with it.

Named the type rather than inlining `Exclude<...>`, since it appears at three
sites:

```ts
// src/models/AppModel/AppLoadState.ts
export type SettableAppLoadState = Exclude<AppLoadState, 'failed'>
```

Confirmed it bites — a throwaway `setLoadState('x', 'failed')` now fails to
compile:

```
error TS2345: Argument of type '"failed"' is not assignable to parameter of type 'SettableAppLoadState'.
```

This was the invariant the branch documented but only enforced by convention, so
the finding is a real tightening: `setLoadFailed` is now the only way into
`'failed'`, which is what keeps a failed row from having a chip and no reason.

`npm run lint` clean; 4261 unit tests pass.

